PR #6, opened 2023-10-31
Doc/update the size of dataset

### LINKED ISSUE #5: Inconsistency in the description
image
The description says that there are 72 collections and 6240084 reviews, but the actual dataset has 71 collections and 4632965 reviews.

L-M-Sherlock:


Expertium:
Now it says:

> Total number of reviews for evaluation: 4,632,965
> 1,606,862 reviews are only used for training.

I don't understand. If that means what I think it means - the standard ML practice of splitting the dataset into "train" and "test" datasets - then isn't that wrong? Unless I completely misunderstand how evaluate.py and script.py work, I'm pretty sure FSRS optimizer is used on every collection, and then evaluate.py calculates weighted metrics based on every collection.
If that's not what you mean, then I'm very confused.

L-M-Sherlock:
I used time series split in the benchmark. 

https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.TimeSeriesSplit.html

The first fold is only used for training. It is not in the test set.

Expertium:
Is it consistent with the built-in Anki implementation?

L-M-Sherlock:
The evaluation modules are different. The models and training processes are consistent.

To test the efficiency of model, we should avoid the information leakage in the evaluation. So I use time series split in the benchmark.

PR #19, opened 2023-12-11
Update analysis to calculate the mode.py

Expertium:


Expertium:
@L-M-Sherlock I finished adding mode estimation, all you have to do is verify that the default parameters that I specified are the ones you used. If they aren't - replace them with the ones you used.

L-M-Sherlock (approved):

PR #22, opened 2023-12-15
Feat/outlier filter & flatter power forgetting curve & new pretrain

L-M-Sherlock:
- [x] Apply outlier filter to other models
- [x] Update FSRS-rs


Expertium:
Just a reminder that this should be v4.5, a new entry. Don't just change v4.
To clarify, if only pretrain was changed, then I would agree that this should be called v4. But since one of the main formulas has been changed, this should be called v4.5.

L-M-Sherlock:
I know about that. Due to a mistake, I have to re-run all the models😭. And I need to compare v4 with v4.5 by using the same outlier filter, so I need to run the script twice.

Expertium:
Minor nitpick: it should be called "FSRS v4.5", not "FSRS-4.5". Since previous versions have the letter v, let's keep it consistent.

L-M-Sherlock:
Due to the semantic version rule, I have used v4.5 in fsrs-optimizer: https://github.com/open-spaced-repetition/fsrs-optimizer/releases/tag/v4.5.0. So it's ambiguous.

analysis.py
Expertium, line 1: I don't think that's a good idea. I've mentioned before that mode is sensitive to clustering. If we don't remove users who have parameters that are very close to their defaults, there will be clusters and the mode will just return the defaults.

L-M-Sherlock: If we check all parameters, the user will be removed even if only one parameter is equal to the default parameter.

Expertium: Yes, but it's better than using a mode which just returns the same default parameters.

L-M-Sherlock: Done.

PR #34, opened 2024-01-12
Change method to bca.py

Expertium:
Initially I set `method` to `percentile`, as it's the most intuitive method. However, after reading "Better Bootstrap Confidence Intervals" (doi:10.1080/01621459.1987.10478410), I came to the conclusion that BCa is the superior method. In retrospect, I should've just left method as default, since scipy defaults to BCa. Anyway, this will require you to update Readme, sorry for making you do extra work. Don't forget to do this in the SM-17 and SM-15 repositories as well.

### DIFF
--- a/evaluate.py
+++ b/evaluate.py
@@ -44,13 +44,13 @@ if __name__ == "__main__":
             values = np.array([item[metric] for item in m])
             identifiers = [i for i in range(len(values))]
             dict_x_w = {identifier: (value, weight) for identifier, (value, weight) in enumerate(zip(values, sizes))}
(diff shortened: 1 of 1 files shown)


L-M-Sherlock:
You can run `python evaluate.py` to get the metrics.

Expertium:
This would take a lot of time. I've never actually ran the benchmark on the full dataset, only on a portion of it. And I can't do this for SM-15 and SM-17. And I don't know how to run it with other models (LSTM, SM-2, etc.).

L-M-Sherlock:
It doesn't cost too much time to run the evaluate code because all results have been stored in the repository.

Expertium:
Ah, ok. I didn't think of that, yeah, I can just copy files from `result` onto my PC.

L-M-Sherlock (approved):

PR #35, opened 2024-01-13
Updated the confidence intervals using method='BCa'.md

Expertium:
You can now merge this too: https://github.com/open-spaced-repetition/fsrs-benchmark/pull/34


Expertium:
Also, please add a link to the main benchmark repo. Ideally, all three repos (main, SM-15 and SM-17) should link to each other

Expertium:
By the way, we're lucky that the confidence intervals are very symmetric. I measure asymmetry like this:
```
            wmean = np.average(values, weights=sizes)
            asymm = (high - wmean)/(wmean - low)
```
If they were highly asymmetric, we would have to report them differently, for example, like this: 1.0 [−0.1, +0.33] or 1.0 [0.9, 1.33].

Expertium:
Don't merge this yet, I will add some minor finishing touches tomorrow. After that everything will be proper: we have confidence intervals that are calculated using the most accurate method available, and results are reported in accordance with the best statistical practices!

Expertium:
By the way, it's now very easy to see how much having a lot of data affects the precision of benchmarking:
image

Expertium:
By the way, you may come across different guidelines for reporting uncertainty:
1) https://sci-hub.se/10.1136/archdischild-2014-307149
2) https://www.bipm.org/documents/20126/2071204/JCGM_100_2008_E.pdf/cb0ef43f-baa5-11cf-3f85-4dcd86f77bd6
3) https://arxiv.org/ftp/arxiv/papers/1301/1301.1034.pdf

When reporting numerical information, our goal is to report as little random noise as possible and only report information that we are certain is "true", in some sense. My rounding function is based on a rather simple idea (which is surprisingly difficult to implement): **_don't report a digit if it could be anything from 0 to 9_**.
For example, 1.5±0.1 means that the last digit could be 4, 5, or 6. It can't be anything else, like 1 or 9. Therefore, we should report it because we are reasonably certain about what the true digit is.
Now consider 1.5±0.5. The last digit could be 0, 1, 2, 3, 4, 5, 6, 7, 8, or 9. In other words, literally anything. Therefore, we should not report it because we are completely uncertain about the true digit.
And we must choose the number of significant digits not just for the mean but for the error as well. My method follows 2 rules: 
1) The number of digits _after the decimal point_ (significant or not) is n for the mean and n+1 for the error. For example, if the mean has 3 digits after the decimal place (including leading zeros), the error should have 4 digits.
2) Two significant digits are reported if the first digit (excluding leading zeros) of the error is 1, 2, 3 or 4, otherwise one significant digit is reported. See examples below to understand how this rule works in practice.

This is why I reported the logloss (ln(number of reviews)) of FSRS v3 as 0.49±0.06, but the logloss of FSRS-4.5 (default parameters) as 0.386±0.0034.
Suppose that the logloss of FSRS v3 is 0.49X±0.06. The last digit of the mean, X, could be off by 6 in either direction. For any X, the [X-6, X+6] interval will contain all ten possible digits. Therefore, we shouldn't report it since we aren't certain about the true digit at all.
But with FSRS-4.5 (default parameters), the situation is different. 0.386±0.0034, the last digit of the mean is 6; it could be off by 3 in one direction (0.38**6**+0.0034=0.38**9**4) and by 4 in the other direction (0.38**6**-0.0034=0.38**2**6). So the true digit could be 2, 3, 4, 5, 6, 7, 8, 9. It can't be 0 or 1. This means that we are at least somewhat certain about the true digit. Therefore, we should report it.

L-M-Sherlock (approved):
LGTM

Expertium:
Idk if you are interested, but I decided to closely examine how often digits change upon resampling.
For example, here are the results for FSRS-4.5, n(reviews), logloss:
```
First digit: {0: 0, 1: 0, 2: 0, 3: 10000, 4: 0, 5: 0, 6: 0, 7: 0, 8: 0, 9: 0}
Second digit: {0: 0, 1: 0, 2: 5182, 3: 4818, 4: 0, 5: 0, 6: 0, 7: 0, 8: 0, 9: 0}
Third digit: {0: 1832, 1: 1435, 2: 839, 3: 484, 4: 262, 5: 272, 6: 604, 7: 975, 8: 1498, 9: 1799}
Fourth digit: {0: 978, 1: 1029, 2: 975, 3: 973, 4: 999, 5: 1014, 6: 950, 7: 1068, 8: 1046, 9: 968}
Most common first digit=3, 100.0%
Most common second digit=2, 51.8%
Most common third digit=0, 18.3%
Most common fourth digit=7, 10.7%
Mean=0.3299
```
Note that I excluded 0, so it's "first digit after the decimal point", not just "first digit", that's always 0. The percentage tells you how often this digit appears. The dictionary countains the number of occurences, for example, `0: 978` means that digit zero occured 978 times out of 10 000 resamples.
As you can see, the first digit doesn't change (it's 3 100% of the time), but other digits can change. The fourth digit is basically random noise, the chances of any digit appearing are roughly equal. Of course, the exact counts change when I re-run the code, since resampled datasets are different every time.

Another example: FSRS-4.5, n(reviews), RMSE (bins):
```
metric: RMSE(bins)
First digit: {0: 10000, 1: 0, 2: 0, 3: 0, 4: 0, 5: 0, 6: 0, 7: 0, 8: 0, 9: 0}
Second digit: {0: 0, 1: 0, 2: 0, 3: 0, 4: 10000, 5: 0, 6: 0, 7: 0, 8: 0, 9: 0}
Third digit: {0: 0, 1: 0, 2: 10, 3: 3520, 4: 6253, 5: 217, 6: 0, 7: 0, 8: 0, 9: 0}
Fourth digit: {0: 1061, 1: 1079, 2: 1035, 3: 1035, 4: 960, 5: 926, 6: 937, 7: 901, 8: 1050, 9: 1016}
Most common first digit=0, 100.0%
Most common second digit=4, 100.0%
Most common third digit=4, 62.5%
Most common fourth digit=1, 10.8%
Mean=0.0442
```
As you can see, we are very certain about the first 3 digits after the decimal place, but the fourth one is just random. This means that my method of rounding digits in the mean and the error (CI) works exactly as I was expecting. If I found that the fourth digit (in this particular case) is not completely random, then I would likely have to re-think my approach.

But the most important part is this:
FSRS-4.5 default parameters, log-loss, ln(n(reviews)). In the repo it's reported as 0.386±0.0034.
```
First digit: {0: 0, 1: 0, 2: 0, 3: 10000, 4: 0, 5: 0, 6: 0, 7: 0, 8: 0, 9: 0}
Second digit: {0: 0, 1: 0, 2: 0, 3: 0, 4: 0, 5: 0, 6: 0, 7: 1, 8: 9997, 9: 2}
**Third digit: {0: 3, 1: 18, 2: 178, 3: 821, 4: 2158, 5: 2978, 6: 2448, 7: 1067, 8: 297, 9: 32}**
Fourth digit: {0: 961, 1: 994, 2: 1010, 3: 971, 4: 1014, 5: 1004, 6: 1019, 7: 972, 8: 1002, 9: 1053}
Most common first digit=3, 100.0%
Most common second digit=8, 100.0%
Most common third digit=5, 29.8%
Most common fourth digit=9, 10.5%
Mean=0.3856
```
Notice that for the third digits, it's very unlikely to be 0, 1 or 9. Their counts are below 1% of the total count=10 000. This means that we can say "The true digit is definitely not 0, not 1 and not 9, we are certain about that". So I was right to report the mean to three digits, the third digit is not random noise.

FSRS v3, log-loss, ln(n(reviews)). In the repo it's reported as 0.49±0.006. Notice that the mean is reported to two digits after the decimal place, but the mean of FSRS-4.5 with default parameters is reported to three digits.
```
First digit: {0: 0, 1: 0, 2: 0, 3: 0, 4: 10000, 5: 0, 6: 0, 7: 0, 8: 0, 9: 0}
Second digit: {0: 0, 1: 0, 2: 0, 3: 0, 4: 0, 5: 0, 6: 0, 7: 14, 8: 8979, 9: 1007}
**Third digit: {0: 568, 1: 414, 2: 415, 3: 620, 4: 1032, 5: 1446, 6: 1672, 7: 1620, 8: 1326, 9: 887}**
Fourth digit: {0: 1008, 1: 949, 2: 965, 3: 1018, 4: 998, 5: 1030, 6: 1001, 7: 1004, 8: 992, 9: 1035}
Most common first digit=4, 100.0%
Most common second digit=8, 89.8%
Most common third digit=6, 16.7%
Most common fourth digit=9, 10.3%
Mean=0.4869
```
Look at the counts for the third digit. None of the counts are below 100, which is 1% of 10 000. In other words, we cannot confidently exclude any digit while saying "That digit is too unlikely to be the true digit". I was right to report the mean to only two digits after the decimal place, the third digit could be anything.
Here's the cool part: my sigdig() function in evaluate.py doesn't need to do resampling every time in order to determine the best way to round the mean and the error (CI), and doesn't need to explicitly count the occurences of digits the way I just showed above. All it needs is the mean and the error (CI).

Here's a text file with analysis of digits of all metrics and all algorithms, in case you are curious:
[Most common digits.txt](https://github.com/open-spaced-repetition/fsrs-benchmark/files/13955964/Most.common.digits.txt)

PR #36, opened 2024-01-17
Add a table with p-values.md

Expertium:


L-M-Sherlock:
Could you provide the code to draw these graphes?

Expertium:
Here
[bar_graphs.zip](https://github.com/open-spaced-repetition/fsrs-benchmark/files/13975696/bar_graphs.zip)
The .py file can be put in the fsrs-benchmark folder, or you can paste the code from it into analysis.py. You should also create a new folder in plots, called bar_graphs.

Expertium:
I have refactored it to remove very similar lines of code and use a function instead.
[bar_graphs.zip](https://github.com/open-spaced-repetition/fsrs-benchmark/files/13977811/bar_graphs.zip)

Expertium:
Btw, I would suggest removing RMSE, for 2 reasons:
1) It's confusing because there is RMSE, and then there is RMSE (bins)
2) We don't care about it. We care about logloss and RMSE (bins)

Expertium:
I renamed this PR, I'll add a table with p-values from paired tests instead. Two tables, actually: for logloss and RMSE (bins). I plan to remove RMSE to make it less cluttered and more concise, I've explained above why we should only keep logloss and RMSE (bins).

Expertium:
Well, I searched high and low, and couldn't find a weighted version of the Wilcoxon signed-rank test. This means that we have 2 options:
1) Use this: https://www.statsmodels.org/dev/generated/statsmodels.stats.weightstats.ttost_paired.html#statsmodels.stats.weightstats.ttost_paired
It's a weighted version of the classic t-test. But it assumes that data is normally distributed. It's also buggy for low p-values - it seems like it just returns 0 for any p-value less than 1e-15.
2) Use scipy.stats.wilcoxon. It doesn't assume normality, but cannot be weighted.

In case you are wondering how much of a difference there is between a t-test and Wilcoxon for our data (without weights), here's a test with FSRS v4 as baseline and FSRS v4 with sqrt(count) in pretrain as comparison:
```
Wilcoxon signed-rank test=3.4465233031965936e-63
Student's paired t-test signed-rank test=3.1514712823011317e-52
```

That's a difference of 11 zeros, or 100 000 000 000 times.

Expertium:
I found another problem. Wilcoxon literally reaches the limit of floating point precision when the sample size and the difference between algorithms is so big. With our data, if I run wilcoxon on every possible pair of algorithms, it can reach 1e-300 with just 2000 collections.
After torturing matplotlib and pandas for hours, this is the best I could do:

Figure_1

Red means "the row algorithm performs worse than the corresponding column algorithm", green is the opposite. Top row and rightmost column are green because FSRS-4.5 outperforms all other algorithms, bottom row and leftmost column are red because HLR performs worse than any other algorithm. I used 1500 collections to prevent Wilcoxon from spitting out zeros due to the limitations of floating point number precision.
Theoretically, it should be possible to make a version that keeps track of the logarithm of p-value rather than the p-value itself, to avoid the issue with floating point precision, but I'm not sure if I can do that. I would have to write my own code for Wilcoxon, and dig really deep into the math.

Expertium:
Nevermind, I found a very good approximation and wrote my own implementation of Wilcoxon. It's quite accurate. For example: -201.51382786924088, -201.5146654571942. The first value is `np.log10(scipy.stats.wilcoxon(data1, data2).pvalue)`, the second value is my approximation. The first 5 digits match, so I would say this is about as good as it gets for all practical purposes.
Now we can use the logarithms of p-values instead of p-values themselves, and then we just need to do a little bit of formatting to convert logarithms to scientific notation.
@L-M-Sherlock I bet you haven't seen p-values this low in your entire life

Figure_2


Of course, with p-values this low, I'm sure a lot of people will ask "Do they even make sense?". Well, Wilcoxon itself uses an approximation for n>50, and I added another approximation to approximate their approximation. But more importantly, Wilcoxon relies on the assumption that the distribution of _pairwise differences_ is symmetrical (even if the data itself doesn't follow a symmetrical distribution). Finally, on the meta-level, any kind of statistical test is just a model of reality and doesn't fully capture all the nuances of the real world. And it definitely doesn't take into account that RMSE depends on the number of reviews. 
Overall, I would say that these p-values make sense on the qualitative level, in other words, we can be sure that all of our algorithms perform differently and that no two algorithms perform exactly the same.

Expertium:
I've removed RMSE. As I explained above, it's confusing because there is RMSE and RMSE (bins), and also we never look at it for comparisons anyway.
Here's the code for making this table. It's a bit messy - you have to run the commented out part first, to save a file.

[significance_table.zip](https://github.com/open-spaced-repetition/fsrs-benchmark/files/14002817/significance_table.zip)

Also, I keep the full FSRS v4 data in a different folder, which is why you will see code like this:
```
        if model == "FSRSv4_old_big":
            name = "FSRSv4"
        else:
            name = model
```
And like this:
```
    for i in range(len(models)):
        if models[i] == 'FSRSv4_old_big':
            models2.append('FSRSv4')
        else:
            models2.append(models[i])
```

Expertium:
@L-M-Sherlock I'm done. Most of these commits are minor changes to the wording. There are only 2 major changes: I added the image with p-values (see .zip file with the code above) and removed RMSE, leaving only two metrics. You can merge this now, and then you can use my plotting code to make a similar table, but for the SM-17 repo.

L-M-Sherlock (approved):
LGTM

PR #37, opened 2024-01-22
Removed misleading phrasing and improved siginificance_table.py.md

Expertium:


Expertium:
@L-M-Sherlock you can merge this

L-M-Sherlock:
Could you update the image?

Expertium:
Done

L-M-Sherlock:
I mean could you replace the image file in your computer and commit the change. I'm running the benchmark so I can't update the image now. Or I will update the image file tomorrow and merge it.

Expertium:
I've updated the image

L-M-Sherlock (approved):

L-M-Sherlock:
Fine. I mean the file: https://github.com/open-spaced-repetition/fsrs-benchmark/blob/main/plots/Wilcoxon-19854-collections.png

But it's OK. I will update the file tomorrow.

Expertium:
I don't understand. I just updated it?
EDIT: ah, you mean not the one in readme, it's a different one

Expertium:
Also, there seems to be an inconsistency between how the saved image looks like and how the `plt.show()` image looks like. I kept changing the images all the time because something was off. And even if I remove `bbox_inches='tight'`, the inconsistency remains. Saving the image manually does not produce the same results as saving it with `plt.savefig()`. Don't be surprised if I make a few more pull requests in this repo and in the SM-17 repo.

PR #38, opened 2024-01-23
Update the image.md

Expertium:
As I mentioned in the previous pull request, I thought that saving an image manually and using plt.savefig() would produce identical results, but no. From now on I will only use plt.savefig().


Expertium:
Sorry for spamming pull requests. Also, I can't edit the image in the plots folder.

L-M-Sherlock (approved):
LGTM

PR #45, opened 2024-01-31
Doc/Mentioned FSRS-4.5 (default parameters) and FSRS-4.5 (only pretrain). (#45)

Expertium:
Mentioned FSRS-4.5 (default parameters) and FSRS-4.5 (only pretrain).


Expertium:
I recommend decreasing the font size of p-values each time a new model is added. The table size is constant, meaning that the size of squares in the table only gets smaller as we add new models. So in order to prevent p-values from overlapping with black borders, the font size needs to be slightly decreased.

L-M-Sherlock (approved):

PR #46, opened 2024-02-06
Add the number of opt. parameters to the table.md

Expertium:


Expertium:
By the way, don't merge this once you're done with the Transformer, there is something else I want you to benchmark: https://github.com/open-spaced-repetition/fsrs-benchmark/issues/47

Expertium:
@L-M-Sherlock I suggest adding the results of Transformer to this PR

PR #52, opened 2024-02-13
Feat/implement ACT-R

### LINKED ISSUE #50: [Feature request] Add the ACT-R model (see paper)
[Relevant paper](http://act-r.psy.cmu.edu/wordpress/wp-content/uploads/2012/12/409s15516709cog0000_14.pdf).
Main formulas:
image
image
The appendix has example calculations.

It only has 4 parameters: decay intercept (a), decay scale (c), activation threshold (𝜏) and noise (s). Seems to be fairly straightforward to implement. In the paper 𝜏 and s are fixed, but I think we should make them optimizable. I don't see why they should be fixed.
Btw, a reminder: please add the Transformer to the table with p-values.

L-M-Sherlock:
waiting for benchmarking results

### DIFF
--- a/other.py
+++ b/other.py
@@ -534,7 +534,7 @@ class ACT_R(nn.Module):
     a = 0.176786766570677  # decay intercept
     c = 0.216967308403809  # decay scale
     s = 0.254893976981164  # noise
-    h = 86400 * 0.025      # time scale
+    h = 86400 * 0.025      # inteference scalar
     tau = -0.704205679427144  # threshold
     init_w = [a, c]
     clipper = ACT_RWeightClipper()
--- a/other.py
+++ b/other.py
@@ -527,6 +527,8 @@ class ACT_RWeightClipper:
             w = module.w.data
             w[0] = w[0].clamp(0, 10)
             w[1] = w[1].clamp(0, 10)
+            w[2] = w[2].clamp(0, 10)
+            w[3] = w[3].clamp(-10, 0)
             module.w.data = w
 
 
@@ -536,7 +538,7 @@ class ACT_R(nn.Module):
     s = 0.254893976981164  # noise
     h = 86400 * 0.025      # inteference scalar
     tau = -0.704205679427144  # threshold
-    init_w = [a, c]
+    init_w = [a, c, s, tau]
     clipper = ACT_RWeightClipper()
 
     def __init__(self, w: List[float] = init_w):
@@ -558,10 +560,10 @@ class ACT_R(nn.Module):
                 )
             )
             m[i] = act
-        return self.activation(m)
+        return self.activation(m[1:])
 
     def activation(self, m):
-        return 1 / (1 + torch.exp((self.tau - m) / self.s))
+        return 1 / (1 + torch.exp((self.w[3] - m) / self.w[2]))
 
     def state_dict(self):
         return list(
@@ -772,7 +774,7 @@ class Trainer:
                 real_batch_size = seq_lens.shape[0]
                 if isinstance(self.model, ACT_R):
                     outputs = self.model(sequences)
-                    retentions = outputs[seq_lens - 1, torch.arange(real_batch_size), 0]
+                    retentions = outputs[seq_lens - 2, torch.arange(real_batch_size), 0]
                 else:
                     if isinstance(self.model, HLR):
                         outputs, _ = self.model(sequences.transpose(0, 1))
@@ -812,7 +814,7 @@ class Trainer:
             real_batch_size = seq_lens.shape[0]
             if isinstance(self.model, ACT_R):
                 outputs = self.model(sequences.transpose(0, 1))
-                retentions = outputs[seq_lens - 1, torch.arange(real_batch_size), 0]
+                retentions = outputs[seq_lens - 2, torch.arange(real_batch_size), 0]
             else:
                 if isinstance(self.model, HLR):
                     outputs, _ = self.model(sequences)
@@ -836,7 +838,7 @@ class Trainer:
 
             if isinstance(self.model, ACT_R):
                 outputs = self.model(sequences.transpose(0, 1))
-                retentions = outputs[seq_lens - 1, torch.arange(real_batch_size), 0]
+                retentions = outputs[seq_lens - 2, torch.arange(real_batch_size), 0]
             else:
                 if isinstance(self.model, HLR):
                     outputs, _ = self.model(sequences)
@@ -888,7 +890,7 @@ class Collection:
             if isinstance(self.model, ACT_R):
                 outputs = self.model(fast_dataset.x_train.transpose(0, 1))
                 retentions = outputs[
-                    fast_dataset.seq_len - 1, torch.arange(len(fast_dataset)), 0
+                    fast_dataset.seq_len - 2, torch.arange(len(fast_dataset)), 0
                 ]
                 return retentions.cpu().tolist()
             else:
--- a/other.py
+++ b/other.py
@@ -534,7 +534,7 @@ class ACT_R(nn.Module):
     a = 0.176786766570677  # decay intercept
     c = 0.216967308403809  # decay scale
     s = 0.254893976981164  # noise
-    h = 86400  # time scale
+    h = 86400 * 0.025      # time scale
     tau = -0.704205679427144  # threshold
     init_w = [a, c]
     clipper = ACT_RWeightClipper()
@@ -548,7 +548,7 @@ class ACT_R(nn.Module):
         :param inputs: shape[seq_len, batch_size, 1]
         """
         m = torch.zeros_like(sp, dtype=torch.float)
-        m[0] = m[0] - torch.inf
+        m[0] = -torch.inf
         for i in range(1, len(sp)):
             act = torch.log(
                 torch.sum(
@@ -557,7 +557,7 @@ class ACT_R(nn.Module):
                     dim=0,
                 )
             )
-            m[i] = m[i] + act
+            m[i] = act
         return self.activation(m)
 
     def activation(self, m):
@@ -1058,10 +1058,10 @@ def process(args):
         p.extend(testset["p"].tolist())
         y.extend(testset["y"].tolist())
 
-    evaluate(y, p, model_name, file)
+    evaluate(y, p, model_name, file, w_list if type(w_list[0]) == list else None)
 
 
-def evaluate(y, p, model_name, file):
+def evaluate(y, p, model_name, file, w_list=None):
     p_calibrated = lowess(
         y, p, it=0, delta=0.01 * (max(p) - min(p)), return_sorted=False
     )
@@ -1081,6 +1081,8 @@ def evaluate(y, p, model_name, file):
         "user": int(file.stem),
         "size": len(y),
     }
+    if w_list:
+        result["weights"] = list(map(lambda x: round(x, 4), w_list[-1]))
     # save as json
     Path(f"result/{model_name}").mkdir(parents=True, exist_ok=True)
     with open(f"result/{model_name}/{file.stem}.json", "w") as f:
--- a/other.py
+++ b/other.py
@@ -518,6 +518,60 @@ class HLR(nn.Module):
         return 0.5 ** (t / s)
 
 
+class ACT_RWeightClipper:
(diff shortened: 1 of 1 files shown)


Expertium:
According to the paper, noise and threshold are fixed. I wonder why, I couldn't find a good explanation. To me it seems like they should be optimizable too.
Of course, if you want to do this exactly as the paper suggests, that's understandable. But I think that making the other two parameters optimizable would make more sense, even if it's not according to the paper.
image

L-M-Sherlock:
> According to the paper, noise and threshold are fixed. I wonder why, I couldn't find a good explanation. To me it seems like they should be optimizable too.

They are untrainable. I try to optimize them, but their gradients are `nan`.

You can check it via following script:

```python
import torch
import torch.nn as nn
from torch import Tensor


class ACT_R(nn.Module):
    a = 0.0176786766570677  # decay intercept
    c = 0.216967308403809  # decay scale
    s = 0.254893976981164  # noise
    h = 86400 * 0.025  # time scale
    tau = -0.704205679427144  # threshold
    init_w = [a, c, s, tau]

    def __init__(self, w = init_w):
        super().__init__()
        self.w = nn.Parameter(torch.tensor(w, dtype=torch.float32))

    def forward(self, sp: Tensor):
        """
        :param inputs: shape[seq_len, batch_size, 1]
        """
        m = torch.zeros_like(sp, dtype=torch.float)
        m[0] = -torch.inf
        for i in range(1, len(sp)):
            act = torch.log(
                torch.sum(
                    ((sp[i] - sp[0:i]) * self.h).clamp_min(1)
                    ** (-(self.w[1] * torch.exp(m[0:i]) + self.w[0])),
                    dim=0,
                )
            )
            m[i] = act
        return self.activation(m)

    def activation(self, m):
        return 1 / (1 + torch.exp((self.w[3] - m) / self.w[2]))

    def state_dict(self):
        return list(
            map(
                lambda x: round(float(x), 4),
                dict(self.named_parameters())["w"].data,
            )
        )


model = ACT_R()

t_history_1 = [0, 1, 2, 4, 8]
t_history_2 = [0, 3, 3, 3, 3]

sp = torch.stack([torch.cumsum(torch.tensor(t_history_1), dim=0), torch.cumsum(torch.tensor(t_history_2), dim=0)], dim=1)

print(sp.shape)

p = model(sp)
print(p)
loss_fn = nn.BCELoss(reduction="none")
loss = loss_fn(p, torch.ones_like(p)).sum()
print(loss)
loss.backward()
print(model.w.grad)
```

Result:
```
torch.Size([5, 2])
tensor([[0.0000, 0.0000],
        [0.9029, 0.8960],
        [0.9473, 0.9435],
        [0.9578, 0.9597],
        [0.9609, 0.9679]], grad_fn=<MulBackward0>)
tensor(200.4808, grad_fn=<SumBackward0>)
tensor([13.1608,  1.6838,     nan,     nan])
```

By the way, ACT-R's time complexity is O(n^2). So it will take days to optimize, which is longer than FSRS.

L-M-Sherlock:
@Expertium, I fixed the `nan` bug. Now the ACT-R model has four trainable parameters.

Edit: Here is the initial results:

```
Model: ACT-R
Total number of users: 238
Total number of reviews: 7300000
Weighted average by reviews:
ACT-R LogLoss (mean±std): 0.328±0.180
ACT-R RMSE(bins) (mean±std): 0.043±0.038

Weighted average by log(reviews):
ACT-R LogLoss (mean±std): 0.380±0.181
ACT-R RMSE(bins) (mean±std): 0.065±0.051

Weighted average by users:
ACT-R LogLoss (mean±std): 0.383±0.182
ACT-R RMSE(bins) (mean±std): 0.069±0.052

weights: [0.0046, 0.25, 0.5368, -0.9978]
Model: FSRS-4.5
Total number of users: 238
Total number of reviews: 7300000
Weighted average by reviews:
FSRS-4.5 LogLoss (mean±std): 0.302±0.165
FSRS-4.5 RMSE(bins) (mean±std): 0.044±0.032

Weighted average by log(reviews):
FSRS-4.5 LogLoss (mean±std): 0.347±0.161
FSRS-4.5 RMSE(bins) (mean±std): 0.063±0.044

Weighted average by users:
FSRS-4.5 LogLoss (mean±std): 0.349±0.162
FSRS-4.5 RMSE(bins) (mean±std): 0.066±0.046
```

It's weird that ACT-R has good performance measured by RMSE(bins), but worse at log loss.

Expertium:
> I fixed the nan bug. Now the ACT-R model has four trainable parameters.

Nice! 

> By the way, ACT-R's time complexity is O(n^2). So it will take days to optimize, which is longer than FSRS.

Yeah, I kinda figured that it would be O(n^2) the moment I saw a for loop inside a for loop in the other user's comment.

Expertium:
> ACT-R RMSE(bins) (mean±std): 0.047±0.044

Wait, this is surprisingly good, what the heck? I did not expect it to perform so well. I thought RMSE will be >10%.

L-M-Sherlock:
I think it's ridiculous that a model which doesn't differentiate the correct response and incorrect response has such a good performance. I plan to draw the calibration graph.

Expertium:
Btw, is FSRS's time complexity O(n)?

L-M-Sherlock:
> Btw, is FSRS's time complexity O(n)?

Of course. FSRS is like RNN. Its each state only relies on the last state. And it's state space is fixed in each step.

L-M-Sherlock:
I know why ACT-R is good at RMSE(bins). The distribution of prediction of ACT-R is very narrow:









As I mentioned before, RMSE(bins) is cheatable. And this is a case that ACT-R cheats RMSE(bins).

Expertium:
Oh, so it's a real-life cheating algorithm that puts all predictions into one bin. I thought we would never see one, unless someone wanted to troll us.

L-M-Sherlock:
Yeah. I plan to reorder the models by log loss after this branch is merged. Otherwise, the ACT-R will be placed in the first rank.

Expertium:
Out of curiosity, try keeping noise and threshold fixed, like in the paper, and make graphs again.

Expertium:
If it turns out that it still cheats, **_DO NOT INCLUDE IT IN THE BENCHMARK_**. Just close as "Unplanned".

L-M-Sherlock:
> If it turns out that it still cheats, DO NOT INCLUDE IT IN THE BENCHMARK. Just close as "Unplanned".

Could you explain why?

Expertium:
Because it's a cheating algorithm. It would make proper algorithms look bad and give readers a false sense of accuracy.

Expertium:
Anyway, first we need to see some calibration graphs with noise and threshold fixed.

L-M-Sherlock:
Metrics of ACT-R with two trainable parameters;

```
Model: ACT-R
Total number of users: 145
Total number of reviews: 4259965
Weighted average by reviews:
ACT-R LogLoss (mean±std): 0.372±0.173
ACT-R RMSE(bins) (mean±std): 0.052±0.040

Weighted average by log(reviews):
ACT-R LogLoss (mean±std): 0.407±0.184
ACT-R RMSE(bins) (mean±std): 0.076±0.056

Weighted average by users:
ACT-R LogLoss (mean±std): 0.409±0.187
ACT-R RMSE(bins) (mean±std): 0.080±0.058

weights: [0.0293, 0.3822]
Model: FSRS-4.5
Total number of users: 145
Total number of reviews: 4259965
Weighted average by reviews:
FSRS-4.5 LogLoss (mean±std): 0.339±0.157
FSRS-4.5 RMSE(bins) (mean±std): 0.047±0.031

Weighted average by log(reviews):
FSRS-4.5 LogLoss (mean±std): 0.361±0.157
FSRS-4.5 RMSE(bins) (mean±std): 0.064±0.041

Weighted average by users:
FSRS-4.5 LogLoss (mean±std): 0.362±0.158
FSRS-4.5 RMSE(bins) (mean±std): 0.067±0.043
```

Calibration graph:

[ACT-R.zip](https://github.com/open-spaced-repetition/fsrs-benchmark/files/14261068/ACT-R.zip)








By the way, I will keep the code merged even if we don't include the results in the benchmark because I will do more research about it in the future.

Expertium:
Some graphs look much better than others. By "better" I mean "less like it's cheating".

Expertium:
@L-M-Sherlock change these:
```
            w[0] = w[0].clamp(0, 10)
            w[1] = w[1].clamp(0, 10)
```
to:
```
            w[0] = w[0].clamp(0.001, 1)
            w[1] = w[1].clamp(0.001, 1)
```

I wonder if maybe our clampings are too generous. None of the values in the paper exceed 1.

L-M-Sherlock:
> I wonder if maybe our clampings are too generous. None of the values in the paper exceed 1.

I have checked the optimized parameters. None of them exceeds [0, 1].

What's your final conclusion? Should I continue the benchmarking?

Expertium:
> What's your final conclusion? Should I continue the benchmarking?
> ACT-R.zip

Some of the graphs look fine, but others look more like cheating. Overall, I'd say it's alright, continue. Oh, and keep the sorting order by RMSE.

Expertium:
Also, out of curiosity, can you tell me how long it takes to run FSRS and ACT-R on, for example, 100 collections? Just to give me a rough idea how much slower ACT-R is.

L-M-Sherlock:
> Also, out of curiosity, can you tell me how long it takes to run FSRS and ACT-R on, for example, 100 collections? Just to give me a rough idea how much slower ACT-R is.

FSRS: 1000 collections per hour
ACT-R: 400 collections per hour

ElementaryZ:
> Some of the graphs look fine, but others look more like cheating. Overall, I'd say it's alright, continue. Oh, and keep the sorting order by RMSE.

Can you explain the calibration graphs and why you say the algorithm is cheating? They don't look like any calibration graphs I'm familiar with.

Expertium:
So here's how it's done:
1) Group predictions into bins, for example: [0.81, 0.82, 0.83, 0.84, 0.85]
2) Group their corresponding real review outcomes into bins as well: [1, 0, 1, 1, 1]
3) Calculate the averages of both: 0.83 and 0.80, respectively
4) Take (average_predicted - average_real)^2
5) Do this for all bins
6) Calculate the weighted root-mean-square error, with the number of reviews within each bin as weight

The idea is that within each bin, the average of predicted probabilities of recall and average retention should be as close as possible. If the algorithm predicts that there is an x% chance that you will get this stuff right, you should get it right x% of the time.
The problem is that this is cheatable: a cheating algorithm can just output a constant that is equal to the average retention of that user. Then all predictions will fall into just one bin. And within that bin, the average of predictions and the average of real outcomes (aka retention) will be exactly the same, achieving an RMSE of 0%. But in the absence of cheats, this metric is not only very useful for evaluating algorithms, but is also intuitively interpretable, unlike the log-loss.

As you can see here, that's pretty much exactly what ACT-R is doing. The distribution of its predictions is very narrow; it concentrates all predictions within a very narrow range.
image

Expertium:
Here's an even better example:

Expertium:
@L-M-Sherlock I've looked at the graphs some more and changed my mind. We shouldn't include ACT-R in the benchmark after all, it cheats too much. Feel free to experiment with it, of course, but don't include the results in readme, in the table with other algorithms.

ElementaryZ:
If it cheats with this specific measure why does it invalidate the log-loss? Have you tried using other measures of fit for binary classification?

Expertium:
It (probably) doesn't invalidate the log-loss, but it's still not a good idea to include an algorithm that basically just outputs a constant every time.
> Have you tried using other measures of fit for binary classification?

We've tried [ICI](https://github.com/open-spaced-repetition/spaced-repetition-algorithm-metric/issues/2). If you can think of anything else, that would be interesting.

Expertium:
@ElementaryZ LMSherlock made some graphs for [DASH](https://github.com/open-spaced-repetition/fsrs-benchmark/issues/51#issuecomment-1942017399), this is what a non-cheating algorithm's calibration graphs look like

Expertium:
Btw, @L-M-Sherlock the graphs are rather cluttered. I would suggest removing weighted least squares and LOWESS, or at least weighted least squares

PR #53, opened 2024-02-13
Update README.md

Expertium:
I added a clarification about the interpretation of average values with different weighting.


L-M-Sherlock (approved):
LGTM

PR #55, opened 2024-02-14
Update significance_table.py

Expertium:
Changed the font size of p-values. Please add DASH. And in the future, please don't forget to decrease the font size of p-values every time you add a new model, and don't forget to update the table.


L-M-Sherlock (approved):
I didn't forget that. The delay is because I want to update it after all benchmarks are complete.

PR #63, opened 2024-02-17
Changed the position of DASH[ACT-R].md

Expertium:
Here's my reasoning:
1) Sort by log loss
2) If there are ties (log loss of algorithm A is equal to log loss of algorithm B **after** rounding), sort them by RMSE. According to this reasoning, DASH[ACT-R] should be above FSRS-rs

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,8 +54,8 @@ The following tables represent the weighted means and the 99% confidence interva
 | Algorithm | Log Loss | RMSE (bins) | Parameters |
 | --- | --- | --- | --- |
 | FSRS-4.5 | **0.32±0.006** | 0.043±0.0010 | 17 |
-| FSRS rs | 0.33±0.006 | 0.046±0.0011 | 17 |
 | DASH[ACT-R] | 0.33±0.005 | **0.039±0.0011** | 5 |
+| FSRS rs | 0.33±0.006 | 0.046±0.0011 | 17 |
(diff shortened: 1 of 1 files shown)


L-M-Sherlock:
If you run the evaluate.py, you will find that FSRS-rs is better than DASH[ACT-R] in log loss.

Expertium:
I specifically said after rounding.

Expertium:
Both have a log loss of 0.33

L-M-Sherlock:
> I specifically said after rounding.

But I think it's correct to rank them before rounding.

Expertium:
The point of rounding is to remove digits that are just random noise. You don't want to rank algorithms based on random noise.

L-M-Sherlock:
But the significant_table thinks it's not random noise.

Expertium:
You mean the table with p-values? It's based on RMSE.

Expertium:
Wait, did you change it? I'm sure it was originally using RMSE.

L-M-Sherlock:
Yeah, I changed it because we rank models by log loss now.

Expertium:
Ok.

Expertium:
@L-M-Sherlock I think the issue is that in the table the results are weighted by n(reviews), but the Wilcoxon test isn't weighted. So the order can be different because of that.

Expertium:
I just realized that I don't know how to update readme if it has been updated between the time when I created this pull request and now.

Expertium:
I guess I'll close this and open another pull request.

PR #66, opened 2024-02-23
Changed the order of some algorithms in the table.md

Expertium:
The problem is that these results are weighted by n(review), but the Wilcoxon test isn't weighted. So the order *can* be different. My logic is that if the log loss is the same, then such algorithms should be sorted by RMSE.


Expertium:
Ah, ok, I found the bug that affected the rounding of ACT-R results. Funnily enough, I can fix it by making the math simpler rather than more complex.

L-M-Sherlock (approved):

PR #71, opened 2024-03-07
Fixed the description of the table with p-values.md

Expertium:


L-M-Sherlock (approved):

Expertium:
Btw, this is a pretty minor thing, but it says "FSRS rs" everywhere except for the table itself, where it says "FSRS-rs", with a dash. Ideally, the namign should be consistent everywhere, either "FSRS rs" or "FSRS-rs".

L-M-Sherlock:
I renamed `FSRS rs` to `FSRS-rs` just now.

PR #79, opened 2024-03-29
Important change to graphing - More representative of actual usage

RlzHi:
Before we directly compared optimiser vs pretrain vs default parameters.
However, sometimes, the full optimiser doesn't have enough data to run but pretrain does.
Instead of ignoring these datapoints, we now default back to pretrain when optimiser fails, and default back to the default parameters when pretrain fails.

image


L-M-Sherlock (approved):
LGTM

PR #80, opened 2024-03-30
Improved tests for: If A is better than B on train, do we use A

RlzHi:
Conclusion: No, the improvement is too small.
It is better to use pretrain for n(reviews)=8 to 63 and full train for 64+

All graphs and stats are contained in this PR


L-M-Sherlock (approved):
LGTM

PR #83, opened 2024-04-07
Update README to mention NN-17.md

Expertium:


Expertium:
Btw, according to Woz, SM-17 has around 30 parameters, so we actually got quite close.

L-M-Sherlock:
Ready to merge?

Expertium:
Yes

PR #99, opened 2024-07-14
Update analysis.py

Expertium:
I realized that KDE uses weighting by sqrt(n(reviews)), while the other two estimates of mode don't, which is inconsistent. Now KDE also assigns equal weights to all datapoints.


L-M-Sherlock (approved):
LGTM

PR #100, opened 2024-07-19
Simplify calculation of sig-digs, handle None values in AUC.py

Expertium:
I decided to simplify my clever method of calculating the number of significant digits, for the sake of keeping it consistent with common practices and not confusing people. I also added code to handle cases where AUC is None.


L-M-Sherlock (approved):
LGTM

PR #104, opened 2024-07-26
Update README.md

Expertium:
Updated one value in the table, fixed an issue caused by a rare edge case in sigdigs.


L-M-Sherlock (approved):
LGTM

PR #106, opened 2024-08-05
decrease font size of significance table

Expertium:
I decreased the font slightly because it's overlapping on the X axis


L-M-Sherlock (approved):
LGTM

Expertium:
You should re-generate the table

L-M-Sherlock:
I'm re-benchmarking, so I will re-generate it later.

PR #107, opened 2024-08-11
GRU-P in bold.md

Expertium:
Since in the "Weighted" table GRU-P is the best according to all three metrics, it makes sense to make it bold.


L-M-Sherlock (commented):
Thanks. LGTM

PR #108, opened 2024-08-11
Update README.md

Expertium:
Forgot to make the number of parameters bold


L-M-Sherlock:
Why do you make the largest number of parameters bold?

Expertium:
So that the entire row is bold

L-M-Sherlock:
I think it's misleading. Given the same accuracy, model with low complexity is better than model with high complexity. If we make the largest number of parameters bold, it may hint high complexity is better.

Expertium:
Yeah, but right now it looks weird because everything in that top row is bold, except for the number of parameters.
If you think that's ok, feel free to close this.

L-M-Sherlock:
OK.

Issue #102, opened 2024-07-24
Inclusion of any of the boosting models


Gilfaro:
Would it be possible to include any of the boosting models like catboost, lightgbm or xgboost, since they are very good with tabular timeseries data such as this and there is already neural network compared.

L-M-Sherlock:
PR is welcome.

Gilfaro:
Will try modifying oither.py as it looks most of the needed preprocessing is already done for other models.

Expertium:
@Gilfaro hey, just wanted to let you know that we're done benchmarking new algos and re-benchmarking old ones, but it's not too late (it's never too late, for that matter) to add more

Gilfaro:
Didn't have much time, but tested boosting on top of fsrs5 as the signals with stability and difficulty.
The results on test data:
```
Model: FSRS-5-catboost
Total number of users: 10
Total number of reviews: 255400
Weighted average by reviews:
FSRS-5-catboost LogLoss (mean±std): 0.3622±0.1144
FSRS-5-catboost RMSE(bins) (mean±std): 0.0418±0.0117
FSRS-5-catboost AUC (mean±std): 0.7254±0.0765

Weighted average by log(reviews):
FSRS-5-catboost LogLoss (mean±std): 0.3505±0.1557
FSRS-5-catboost RMSE(bins) (mean±std): 0.0528±0.0221
FSRS-5-catboost AUC (mean±std): 0.7213±0.0784

Weighted average by users:
FSRS-5-catboost LogLoss (mean±std): 0.3480±0.1606
FSRS-5-catboost RMSE(bins) (mean±std): 0.0545±0.0233
FSRS-5-catboost AUC (mean±std): 0.7189±0.0757
```
```
Model: FSRS-5
Total number of users: 10
Total number of reviews: 255400
Weighted average by reviews:
FSRS-5 LogLoss (mean±std): 0.3737±0.1183
FSRS-5 RMSE(bins) (mean±std): 0.0496±0.0188
FSRS-5 AUC (mean±std): 0.7064±0.0897

Weighted average by log(reviews):
FSRS-5 LogLoss (mean±std): 0.3660±0.1512
FSRS-5 RMSE(bins) (mean±std): 0.0647±0.0245
FSRS-5 AUC (mean±std): 0.6741±0.1350

Weighted average by users:
FSRS-5 LogLoss (mean±std): 0.3641±0.1547
[...]

PR #109, opened 2024-08-18
WIP: use Catboost on top of FSRS-5

Gilfaro:
#102
Sample usage of Catboost on top of FSRS-5.
It uses "delta_t", "power_forgetting_curve", "stability", "difficulty" from FSRS-5 as input to the boosting algorithm.

### DIFF
--- a/script-catboost.py
+++ b/script-catboost.py
@@ -1,0 +1,459 @@
+import os
+import sys
+import json
+import numpy as np
+import pandas as pd
+from tqdm import tqdm
+from pathlib import Path
+import matplotlib.pyplot as plt
+from sklearn.model_selection import TimeSeriesSplit
+from sklearn.metrics import root_mean_squared_error, log_loss, roc_auc_score
+from statsmodels.nonparametric.smoothers_lowess import lowess
+from concurrent.futures import ProcessPoolExecutor, as_completed
+from itertools import accumulate
+import torch
+from catboost import CatBoostClassifier, Pool
+
+dev_mode = os.environ.get("DEV_MODE")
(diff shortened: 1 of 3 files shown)


script-catboost.py
```diff
@@ -0,0 +1,459 @@
...
+        )
+
+        model = CatBoostClassifier(
+            verbose=False, allow_writing_files=False, has_time=True
+        )
+        f = model.fit(train_pool, eval_set=eval_pool, early_stopping_rounds=100)
+        # print(f"Model accuracy on eval data: {f.score(eval_pool):.5%}")
+        # print(f.get_feature_importance(prettified=True))
+
+        prob = f.predict_proba(eval_pool)
```
L-M-Sherlock, line 106: It's not very useful for production if we predict the probability directly because the final result is the interval scheduled for the user. How can we find out the inverse function of it?

Gilfar: For predicting interval you would need to build a completely different booster. You would swap delta_t and y.
Then probably would be good idea to change y to rating if you want all difficulty buttons and maybe delta_t for the last correct delta_t.

Gilfaro: For predicting interval you would need to build a completely different booster. You would swap delta_t and y.
Then probably would be good idea to change y to rating if you want all difficulty buttons and maybe delta_t for the last correct delta_t.

Expertium:
~~Let's include HLR-catboost instead~~
I have a much better idea: https://github.com/open-spaced-repetition/srs-benchmark/issues/102#issuecomment-2296640088

PR #110, opened 2024-08-20
use FSRS4.5+ forgetting curve for algos that do not have implicitly defined curve

Gilfaro:
Use FSRS4.5+ forgetting curve for algos that do not have implicitly defined curve.
This improves all of the NN based algos, but you will need to pretrain again.
Similar thing could be done for HLR which improves it as well, but it wouldn't be half life anymore, not sure if you would want to have 2 variants to see how forgetting curve alone impacts the same algo.

### DIFF
--- a/other.py
+++ b/other.py
@@ -787,7 +787,7 @@ class RNN(nn.Module):
         return self.fc(h)
 
     def forgetting_curve(self, t, s):
-        return 0.9 ** (t / s)
+        return (1 + FACTOR * t / s) ** DECAY
 
 
 class GRU_P(nn.Module):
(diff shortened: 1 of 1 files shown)


Expertium:
Yeah, let's not touch HLR, since that's how it was originally defined. No problem with adding the new curve to GRU and Transformer, though.
EDIT: and don't change NN-17 either, it's meant to be similar to [SM-17](https://supermemo.guru/wiki/Algorithm_SM-17).

other.py
```diff
@@ -875,7 +875,7 @@ def forward(self, src):
         return output, None
 
     def forgetting_curve(self, t, s):
```
L-M-Sherlock, line 877: This `forgetting_curve` is unused in the prediction of `GRU_P`.

Gilfaro: That is correct, it doesn't even have such function, only Transformer and RNN classes are modified

L-M-Sherlock (approved):
LGTM

L-M-Sherlock:
I will re-run the benchmark soon.

PR #113, opened 2024-08-29
A new way to visualize performance.py

Expertium:
I added a new table to visualize performance of each algorithm. It shows the % of users, for whom algorithm A (row) has lower RMSE than algorithm B (column). For example, 95% means that for mor users, algorithm A results in lower RMSE, and only for a small minority of users, algorithm B results in lower RMSE.
I'll also edit the README.

### DIFF
--- a/superiority.py
+++ b/superiority.py
@@ -1,0 +1,152 @@
+import json
+import math
+import pathlib
+import warnings
+import matplotlib
+import matplotlib.pyplot as plt
+from matplotlib import cm
+from matplotlib.colors import ListedColormap, LinearSegmentedColormap
+import numpy as np
+import pandas as pd
+from scipy import stats
+
+warnings.filterwarnings("ignore")
+
+if __name__ == "__main__":
+    models = [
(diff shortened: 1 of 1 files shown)


Expertium:
Okay, well, I don't know how to modify multiple files in the same PR
https://github.com/open-spaced-repetition/srs-benchmark/pull/114

Expertium:
And I don't know how to un-draft this either...

L-M-Sherlock (approved):
It's very useful!

Expertium:
I just realized that there is a problem. If in algo1.jsonl the data is like this:
```
user1: 0.5
user2: 0.4
user3: 0.3
user2: 0.2
```

But in algo2.jsonl the data is like this:
```
user1: 0.5
user3: 0.3
user3: 0.4
user2: 0.2
```

The pairwise comparison will be inaccurate. In other words, everything has to be sorted by `user`. I'm currently trying to fix it, I'll either make another PR or ask for your help later.

PR #115, opened 2024-08-29
Ensure that data is sorted by 'user'.py

Expertium:


L-M-Sherlock:
Does it change the superiority table image?

Expertium:
It doesn't. Maybe I was worried about sorting for no reason. Still, it's nice to have an extra sanity check.

L-M-Sherlock (approved):
LGTM

PR #116, opened 2024-08-29
Ensure that data is sorted by 'user'.py

Expertium:
I'm not 100% sure that your .jsonl output is guaranteed to be in order, so I added this.

### DIFF
--- a/significance_table.py
+++ b/significance_table.py
@@ -133,24 +133,28 @@ if __name__ == "__main__":
     sizes = []
     for model in models:
         print(f"Model: {model}")
-        RMSE = []
-        logloss = []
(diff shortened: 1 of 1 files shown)


L-M-Sherlock:
The jsonl data is sorted by user id. Which order do you want to sort?

Expertium:
By user ID. I'm just not sure if it always works. I think when I was running other.py locally, sometimes the order would break.
If you think these changes are redundant, close this PR. Also, please read what I said in Discord, in DMs.

L-M-Sherlock:
> I think when I was running other.py locally

I guess you interrupted the benchmark code during training. The sorting is done in the end of benchmarking.

PR #118, opened 2024-08-31
Update README.md

Expertium:
SM2 -> SM-2
Also, according to Discord users, "dry run" is unclear, so I changed it. And I amde some cosmetic changes to make thigns look slightly nicer


L-M-Sherlock (approved):
LGTM

PR #119, opened 2024-09-03
Deal with rounding errors in superiority.py

Expertium:
Sometimes, as a result of rounding, the values in some cells don't add up to 100.0%, and add up to 100.1% instead. Example:
image

I fixed that. This change requires re-generating the diagram.


L-M-Sherlock:
Could you update the image?

Expertium:
I can update it in README, but you'll have to add it to `plots` yourself.

L-M-Sherlock:
Fine. I will add it tomorrow.

PR #120, opened 2024-09-04
Updated the superiority image.md

Expertium:
I can never figure out how to make changes to multiple files in the same PR. I've googled it, and I still have no clue.


L-M-Sherlock:
Don't worry. I will add the image later.

PR #122, opened 2024-09-13
No log scale

Expertium:
I've said this before, but I think using the log scale makes it harder to tell the shape of the distribution, and which values of a parameter are common and which aren't.
This requires re-generating all histogram plots of parameters.


L-M-Sherlock:
w 0
w 1
w 2
w 3
w 4
w 5
w 6
w 7
w 8
w 9
w 10
w 11
w 12
w 13
w 14
w 15
w 16
w 17
w 18

PR #135, opened 2024-12-03
Mention FSRS-5 presets

Expertium:


Expertium:
Btw, add FSRS-5 to the images, too

L-M-Sherlock (approved):
LGTM

L-M-Sherlock:
> Btw, add FSRS-5 to the images, too

I will add it after the benchmark of FSRS-5-deck is finished.

Expertium:
In that case I recommend also waiting for FSRS-5S (siblings), which I will make once I have the necessary dataset and modified optimizer code so that I can use `sibling_review`

L-M-Sherlock:
Duplicated of #139

PR #140, opened 2024-12-18
Use only three threads per process

Expertium:


Expertium:
From Discord:

> @Jarrett Ye not trying to be mean or anything, but I really wish you made the benchmark code in a way that it actually uses a fixed number of threads
> This has been a problem for a long time, for as long as I can remember running the benchmark code, actually.
> I have a 6-cores, 12-threads CPU (Ryzen 5 3600), and this is what happens when I run other.py with 2 threads:
image

> It should be around 16-17%
> 2/12
> Sometimes it goes as high as 40%
> ...or even higher
image

> And this is with 3 threads, which should be at most 3/12=25%
image

Rossgb:
> I had a quick look and it seems he is limiting only processes not threads. I have no idea what I am doing and this may go horribly wrong but try putting this directly after the imports: 

```
torch.set_num_threads(1)
torch.set_num_interop_threads(1)
```

I tried it, and it fixed the problem. Now the CPU usage is as expected.

Expertium:
Actually, hold on, don't merge this yet. If I use a high number of threads, I get an error and the code crashes for some reason
I can't screenshot it, but it's some memory-related error, something like this (this is just an image I got from Google, but the message is similar):
image

Also, all 32 gigs of RAM end up being used

Expertium:
Alright, I have experimented with different ratios. 2-3 threads per process seems good. Let's make it 3.

Expertium:
Rename "threads" to "processes" and mention that it's 3 threads pre process.

L-M-Sherlock (approved):
LGTM

PR #145, opened 2024-12-30
Changed boldness in the table

Expertium:
I want to discuss this a bit. Right now we don't have a clear guideline on how to highlight the best result. We could use one of these two rules:
1) If an algorithm outperforms all other algorithms on all 3 metrics (aka it's a clear winner), highlight its name. If it only outperforms other algorithms on 1-2 metrics, do not highlight **any** algorithm names.
2) If an algorithm outperforms all other algorithms on all 3 metrics, highlight its name. If it only outperforms other algorithms on 1-2 metrics, highlight the name of every algorithm that is the best according to at least one metric. This means that up to 3 algorithm names can be highlighted.

Regarding the current case, in the unweighted table FSRS-5 recency outperforms GRU-P-short. If we go by rule 1, we shouldn't highlight it, and we shouldn't highlight GRU-P-short either. If we go by rule 2, we should highlight both.

@brishtibheja @user1823 bikeshedding time!


brishtibheja:
I was thinking about this today but the only good way to help people seemed having the same table on a website where the same table is interactive so people can change sorting etc.

The **easiest** way to do this is create a table in notion and make it available on Web, and then we just link from here. Wdyt?

Expertium:
Idk, I'd rather keep it simple

user1823:
I think that rule 2 is better.

Expertium:
Yep, I think so, too. Let's see what Jarrett thinks.

L-M-Sherlock:
I'm OK with both.

Expertium:
Nice. Let's merge it then

Expertium:
@L-M-Sherlock it's ready to be merged

Expertium:
And btw, I changed the names a little bit, so don't forget to update them in superiority.py and elsewhere too
EDIT: I did it myself

L-M-Sherlock (approved):
LGTM

PR #146, opened 2025-01-01
Updated the comment under the superiority table

Expertium:


Expertium:
Actually, are spaces arouhd `\n` necessary?
`"Anki-SM-2 \n trainable"`

I don't think they are.

L-M-Sherlock (approved):
LGTM

Issue #137, opened 2024-12-10
TODO list


Expertium:
This is just to keep track of stuff
1) ~~Add sibling information in a way that FSRS can work with. https://github.com/open-spaced-repetition/srs-benchmark/issues/136#issue-2718136827, https://discord.com/channels/368267295601983490/1282005522513530952/1320698604771348570~~ Done ✅
2) ~~During testing, filter out same-day reviews using delta_t in days, but for calculations keep delta_t in seconds. https://forums.ankiweb.net/t/due-column-changing-days-from-whole-numbers-to-decimals-in-scheduling/52213/53?u=expertium~~
Done ✅
3) ~~Benchmark obezag's idea. https://discord.com/channels/368267295601983490/1282005522513530952/1315873110171451422~~ ❌
4) ~~Fine-tune the formula for interpolating missing S0: https://discord.com/channels/368267295601983490/1282005522513530952/1319714323680989256~~ ❌
5) ~~Benchmark updating D before S~~ Done ✅
6) ~~Benchmark setting weights of outliers to 0: https://discord.com/channels/368267295601983490/1282005522513530952/1320670294188363778~~ ❌
7) ~~Benchmark FSRS v1 and FSRS v2~~ Done ✅

L-M-Sherlock:
What's "Fine-tune the formula for interpolating missing S0"?

Expertium:
Check the Discord message, I attached a link

> 1) In the 10k dataset, find users who haven't used one or two buttons during the first review, but started doing so in the second half of their review history
> 2) Estimate the missing S0 https://github.com/open-spaced-repetition/fsrs-optimizer/blob/9b9b700ea463a2505f28d8c04717d9bd34787d5e/src/fsrs_optimizer/fsrs_optimizer.py#L1107 based on the first halves of their review histories
> 3) Run the optimizer on the second halves, don't calculate S0 normally, fill in S0 using the previous estimates. Example: a user never used Good. You use your wacky formula (that I don't understand) to estimate S0. Then, for the second half of the review history, one where the user does use Good, use your estimate from the previous half
> 4) Get logloss/RMSE values, tweak w1 and w2 in your wacky formula
> 5) Repeat steps 2-4 many times until you find the best combination of w1 and w2

Expertium:
Also, don't forget to benchmark FSRS v1 and FSRS v2. I'll add that to the list

L-M-Sherlock:
[...]

PR #148, opened 2025-01-04
Fine-tuned w1 and w2

Expertium:
See https://github.com/open-spaced-repetition/srs-benchmark/issues/137#issuecomment-2571384289

Of course, I want you to implement this in the optimizer.

### DIFF
--- a/other.py
+++ b/other.py
@@ -193,8 +193,8 @@ class FSRS(nn.Module):
                     else:
                         rating_stability[small_rating] = rating_stability[big_rating]
 
-        w1 = 1.35
-        w2 = 0.68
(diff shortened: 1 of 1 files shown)


L-M-Sherlock (approved):
LGTM, I will re-benchmark FSRS-5-pretrain tomorrow.

Expertium:
I don't think it will affect the metrics. Interpolation is needed only if a button or two have never been used during the first review. But if those buttons have never been used, that means their respective S0 will also never be used!
In other words, we cannot see any difference using our current benchmarking setup. But there will be a difference for those real users who have never used a button and are now starting to use it.

PR #162, opened 2025-01-20
Fixed --equalize_test_with_non_secs after recent commit

1DWalker:
The recent [commit](https://github.com/open-spaced-repetition/srs-benchmark/commit/0c3a2c818140fd52459792103d84a836530cb081) does a rename on df to df_secs but misses the return. This PR fixes it.


L-M-Sherlock (approved):
Thanks!

Issue #161, opened 2025-01-20
Make a shorter table and link to an external database


Expertium:
https://discord.com/channels/368267295601983490/1282005522513530952/1330993680735014952

> obezag — Today at 23:13
> Just to stir the pot a little bit more...
> 
> @Expertium 
> 
> Maybe it makes sense to only have two small representative tables displayed inline in the Readme. 
> E.g. consisting of only the "best" of each "family" of scheduler: e.g. only FSRS-5-recency (for FSRS) and GRU-P short (for GRU-P).
> 
> Then have a link to some sort of externally hosted, easily filterable / sortable table (e.g. a Google sheet) that contains every algorithm, their stats, and what additional input features they use. 
> 
> That way the display table gives a quick overview of the SOTA for SRS algorithms, and the external sheet can be drilled down into to with filters to compare different algorithms in different ways, e.g. just all --shorts or just all "uses sibling information" etc. etc.
> (Unfortunately, it doesn't look like you can embed anything fun like a google sheet into your Github readme.)
> (So it would just be a link.)

PR #167, opened 2025-01-25
Add the LSTM model

1DWalker:


Expertium:
I thought you were going to use more input features, such as the time of the day and the number of reviews done on that day before the current review (this stuff: https://imgur.com/a/Qrl5tEV).

Anyway, I think it's time we add a column for input features.

1DWalker:
I mentioned that I tried those features but they increased train accuracy but decreased validation accuracy. It just helped the model to overfit to each user a little. If I try that again then I would need more regularization or I would train on more users first. I've left some of the code that creates those features in the `create_features_helper` function in case anyone wants to use it in the future.

@L-M-Sherlock thoughts on additional columns in the README tables to indicate which features are used by which model? It is related to https://github.com/open-spaced-repetition/srs-benchmark/issues/161.

L-M-Sherlock:
The README is too long now. I recommend recording it in wiki.

L-M-Sherlock:
Do you mind merging it at first?

1DWalker:
I'm not sure what you mean by that. There should be no conflicts with the main branch I think.

L-M-Sherlock:
I mean we can improve the document in the next PR or build the wiki. This PR looks good to me now.

PR #168, opened 2025-01-26
Added a new column

Expertium:
I added a column with input features. Ngl, I feel like my idea of using abbreviations isn't that great. But typing all that out would make the table huge, and adding more columns isn't great either, so I guess using a ton of abbreviations is the lesser evil.
@1DWalker idk about the RMSE bins exploit thingy, do we need to specify some input features? I mean, it uses interval lengths and grades...I think?


1DWalker:
I think RMSE bins exploit can be considered "IL+G"

Expertium:
Do you think this column should be moved to the left, between "Log Loss" and "Parameters"?
Or maybe do the opposite and move "Parameters" to the far right?

1DWalker:
I think it's fine as it is.

L-M-Sherlock:
Please add them to here as well: https://github.com/open-spaced-repetition/srs-benchmark/blob/0cad52872832e982a267245ad3b3c93ec407287b/evaluate.py#L90-L128

Expertium:
Input features? Why?

Expertium:
@L-M-Sherlock this PR is ready

L-M-Sherlock (approved):
LGTM

PR #174, opened 2025-02-11
Create superiority_small.py and significance_table_small.py

### LINKED ISSUE #161: Make a shorter table and link to an external database

Expertium:
From now on let's use this table in the README, and add a link to the full table


Expertium:
Here's what it looks like: 

Superiority-9999

Expertium:
Wilcoxon-9999-collections

Expertium:
Btw, we **really** should ask Dae to give us one more collection to replace that one guy with not enough reviews. This would:
1) Finally bring the number of collections to exactly 10,000, just like it says
2) Make it less likely that something will crash during optimization by eliminating this particular failure point

L-M-Sherlock:
> From now on let's use this table in the README, and add a link to the full table

Could you update the README.md in this PR?

Expertium:
@L-M-Sherlock I updated the README, but I don't know how to link to an image without it, well, appearing as an image. Hope you can help.
image

L-M-Sherlock (approved):
LGTM

PR #181, opened 2025-02-14
Don't load unused columns

Luc-Mcgrady:
(Also moves the "rating" drop into the read_parquet filter)
Seems to help with memory usage:

column filter
filter columns

Can't filter `elapsed_days` on secs because of:
https://github.com/open-spaced-repetition/srs-benchmark/blob/b6ae34dcff0146a2a2af192a177964b784d88f26/script.py#L270-L271

--secs:
--secs

Tested with: https://github.com/Luc-Mcgrady/srs-benchmark/blob/pq-filters-test/performance.py


L-M-Sherlock (approved):
LGTM

PR #182, opened 2025-02-15
Added a taxonomy of algorithms

Expertium:
I systematized the algorithms based on their architecture. Also, the usage of "algorithm" and "model" was inconsistent, so I made it more consistent.
To make it even more consistent, Jarrett, I suggest replacing `--model` with `--algo`


Expertium:
From now on, let's refer to the specific implementation as "algorithm" and to the high-level mathematical model that it is based on as "model". So DSR is a "model", but FSRS is an "algorithm".

L-M-Sherlock:
Should we also change this?

https://github.com/open-spaced-repetition/srs-benchmark/blob/b6ae34dcff0146a2a2af192a177964b784d88f26/config.py#L68

Expertium:
Yes, that's what I was talking about. Replace `--model` with` --algorithm` or just `--algo`

L-M-Sherlock (approved):
LGTM

PR #184, opened 2025-02-24
Remove the Transformer

Expertium:
According to @1DWalker the current implementation is likely bugged


Expertium:
@1DWalker, do you think we should remove GRU and GRU-P right now, or is it too early?

Btw, I don't want to remove LSTM in the future, since then we will only have one neural network, RWKV-7. I think we should have at least 2 neural nets. So I think it's fine to remove GRU and GRU-P and keep LSTM forever.

1DWalker:
Since you want 2 neural networks, we could keep GRU-P for now and remove GRU. GRU-P has some significance for being the only nn model that predicts recall directly.

L-M-Sherlock:
RWKV-7 is too large for our case. And I want to keep GRU because my paper proposed GRU-HLR.

1DWalker:
What would be a parameter count that would work? I'm not training a model with billions of parameters like for LLMs, I'm working on a model with ~1 million parameters.

L-M-Sherlock (approved):
LGTM

PR #185, opened 2025-03-07
64-bit precision when calculating weighted variance

Expertium:
Btw, I don't think the weights are normalized to add up to 1, that comment is incorrect


L-M-Sherlock (approved):
LGTM

PR #143, opened 2024-12-28
Expt/trainable grade

L-M-Sherlock:

PR #187, opened 2025-03-13
Expt/decay as a function of D

L-M-Sherlock:
port from https://github.com/open-spaced-repetition/fsrs-optimizer/pull/168

to benchmark it, run this command:

```bash
python other.py --algo FSRS-D
```


Expertium (approved):

Expertium:
Can you also add trainable grades, like here: https://github.com/open-spaced-repetition/srs-benchmark/pull/143

No need to benchmark it, just add the code

L-M-Sherlock:
I won't merge this branch into the main branch because it contains too many verbose code. This PR is just for testing until the new idea is proved better and we decide to implement it.

For the trainable grade... I will consider it later.

Gilfaro:
There might be some error in the monotonic implementation as always w[6] < w[7] and it should be w[4] > w[5] > w[6] > w[7] from my understanding
Also I tested on a subset of the dataset with ```DECAY = -0.05 * d``` and you already can get up to ~3-8% relative improvement in RMSE even without the monotonicity fix

Expertium:
Yes, one of the things I want to test is a better mapping between D and decay. But I want Jarrett to fix monotonicity first.
Btw, Jarrett. I wonder, why don't we do this?
```
            w[0] = w[0].clamp(S_MIN, 100)
            w[1] = w[1].clamp(w[0], 100)
            w[2] = w[2].clamp(w[1], 100)
            w[3] = w[3].clamp(w[2], 100)
            w[4] = w[4].clamp(1, 10)
            w[5] = w[5].clamp(w[4], 10)
            w[6] = w[6].clamp(w[5], 10)
            w[7] = w[7].clamp(w[6], 10)
```
Clamp one parameter based on the other one. This guarantees monotonicity. Idk if it does anything weird to the optimizer though.

L-M-Sherlock:
> I wonder, why don't we do this?

If the user don't use `hard`, w[1] and w[5] both are fixed during the optimization, so w[2] and w[6] cannot be less than them.

Expertium:
Alright, I guess I'll use this code as is, screw monotonicity

Expertium:
So I'm trying different hyperparameters for this curve that maps D to decay: https://www.desmos.com/calculator/biojzdd1nv
Like this:
```
    def forgetting_curve(self, t, s, d):
        a1 = 1.6
        a2 = 4.3
        a3 = -0.65
        a4 = -0.07
        e = 2.718281828459045
        transformed_d = a1 * (d - a2)
        exp = e ** -transformed_d
        DECAY = a3 / (1 + exp) + a4
        FACTOR = 0.9 ** (1 / DECAY) - 1
        return (1 + FACTOR * t / s) ** DECAY
```
I'm getting really strange results. Log-loss weighted by n(reviews) is (so far) always worse, but RMSE weighted by n(reviews) is sometimes better, and log-loss weighted by n(users) is sometimes better, and RMSE weighted by n(users) is sometimes better. So for some incomprehensible reason only log-loss weighted by n(reviews) is always worse.
image

This suggests that this change benefits people with a small number of reviews, but not people with a lot of reviews.
Also, I don't know how the hell it is statistically significantly **worse** if log-loss weighted by users is **better**.

I'll try 16 different combinations of hyperparameters a1-a4. If I manage to find something better than FSRS-5, I'll try making  grades optimizable - something that I always thought must work - to improve it further. If no combination of hyperparameters provides an improvement over FSRS-5, then this idea is not good.

Gilfaro:
For most users the best fit is somewhere between (-0.5,0) for the exponent so your function is way too steep.
This is why the closer you get to -0.05x the better results you get in both metrics.

Expertium:
I tried 17 combinations of hyperparameters, all sucked

image

So I suppose we can close this PR

PR #190, opened 2025-03-23
FSRS-6

L-M-Sherlock:
code: https://github.com/open-spaced-repetition/fsrs-optimizer/pull/169 & https://github.com/open-spaced-repetition/fsrs-rs/pull/313


Expertium:
I recommend experimenting further to see if you can add some sort of _f_(D) term as well, like in the two main formulas

L-M-Sherlock:
@Expertium it doesn't work well:
- https://github.com/open-spaced-repetition/srs-benchmark/pull/193

Expertium:
Make sure that the superiority and statistical significance images use show **the same** algorithms, as right now they are inconsistent. Remove RMSE-BINS-EXPLOIT and SM-2, use Anki-SM-2 with default parameters instead. Remove FSRS v3, replace FSRS-5-recency with FSRS-5 and add FSRS-6-recency instead
So the algorithms in **both** tables should be: LSTM, FSRS-6-recency, GRU-P-short, FSRS-5, FSRS-4.5, FSRS v4, DASH, NN-17, HLR, ACT-R, Ebisu and Anki-SM-2 with default parameters.

L-M-Sherlock:
> Make sure that the superiority and statistical significance images use show **the same** algorithms, as right now they are inconsistent

Sorry, I cannot find the inconsistency. Where did you find the inconsistency?

Expertium:
image
Look at which algorithms are present in the Superiority table and in the statistical significance table. They must be the same.
RMSE-BINS-EXPLOIT is present only in one table. Also, in one table you use Anki-SM-2, but in the other one you use Anki-SM-2 with default parameters.


> Remove RMSE-BINS-EXPLOIT and SM-2, use Anki-SM-2 with default parameters instead. Remove FSRS v3, replace FSRS-5-recency with FSRS-5 and add FSRS-6-recency instead
> So the algorithms in both tables should be: LSTM, FSRS-6-recency, GRU-P-short, FSRS-5, FSRS-4.5, FSRS v4, DASH, NN-17, HLR, ACT-R, Ebisu and Anki-SM-2 with default parameters.

Expertium:
Unrelated, but I suggest adding FSRS-6 to other.py, to make my life easier if I ever decide to tweak it

L-M-Sherlock:
> Look at which algorithms are present in the Superiority table and in the statistical significance table. They must be the same.
> RMSE-BINS-EXPLOIT is present only in one table. Also, in one table you use Anki-SM-2, but in the other one you use Anki-SM-2 with default parameters.

I guess you checked the old table instead of the new table in of this PR:

image

> Unrelated, but I suggest adding FSRS-6 to other.py, to make my life easier if I ever decide to tweak it

I added it in the latest commit.

Expertium:
```python
    trainer = Trainer(
        model,
        df,
        None,
        n_epoch=32,
        lr=4e-2,
        wd=1e-4,
        batch_size=65536,
    )
```
replace it with something like
```python
    trainer = Trainer(
        model,
        df,
        None,
        n_epoch=5 if MODEL_NAME == "FSRS-6" else 32,
        lr=4e-2,
        wd=1e-5 if MODEL_NAME == "FSRS-6" else 1e-4,
        batch_size=512 if MODEL_NAME == "FSRS-6" else 65536,
    )
```

Expertium:
image
image

wd is 1e-4 in some places and 1e-5 in other places

L-M-Sherlock:
> wd is 1e-4 in some places and 1e-5 in other places

The weight decay will make the parameters close to zero, which makes sense in neural network. But it is not appropriate for FSRS.

Expertium:
Yes, but I screenshotted wd specifically for FSRS

L-M-Sherlock:
The wd is not used by the Trainer for FSRS, so it doesn't matter.

PR #192, opened 2025-03-24
add HighProbabilityBinomialApproximatorCDF

### LINKED ISSUE #189: Implement "historical likelihood" in FSRS
After beating Claude with a stick multiple times I just barely managed to make it write a fully "torched" version of my Poisson binomial approximation
It still has issues though: unused variables and one last "if" instead of "torch.where". No matter how many times I try, it always either has unused variables or one "if", I just can't get it to finish this properly.

I really hope you can implement this directly in FSRS itself. In other words, if Claude did 80% of the work, I hope that you can do the remaining 20%.

Torch implementation that takes summary statistics as inputs:
https://drive.google.com/file/d/1bku1aG_IJty0HhWATLvwwS6YRZHtHZ-m/view?usp=sharing

Numpy implementation that takes probabilities as inputs:
```
class HighProbabilityBinomialApproximatorCDF:
    def __init__(self):
        self.sum_p = 0.0
        self.sum_p_squared = 0.0
        self.sum_log_q = 0.0  # For the boundary case
        self.n = 0

    def add_trial(self, p):
        """Add a new Bernoulli trial with probability p"""
        self.sum_p += p
        self.sum_p_squared += p * p
        self.sum_log_q += np.log(1 - p) if p < 1 else float('-inf')
        self.n += 1

    def get_heterogeneity(self):
        """Calculate coefficient of variation of probabilities"""
        if self.n <= 1:
            return 0.0

        p_avg = self.sum_p / self.n
        p_variance = (self.sum_p_squared / self.n) - (p_avg * p_avg)
        p_std_dev = np.sqrt(max(0, p_variance))  # Avoid numerical issues

        # Coefficient of variation
        return p_std_dev / p_avg if p_avg > 0 else 0

    def cdf(self, k):
        """Calculate probability that number of successes is <= k"""
        # Handle boundary cases
        if k < 0:
            return 0.0
        elif k == 0:
            return np.exp(self.sum_log_q)  # Product of (1-p_i)
        elif k == self.n:
            return 1.0

        # Get average probability and heterogeneity
        p_avg = self.sum_p / self.n
        heterogeneity = self.get_heterogeneity()

        # Parameters that can be fine-tuned (same as in PMF)
        coeff1 = 0.0256
        coeff2 = 0.0200
        coeff3 = 0.95
        thresh = 0.98

        # For highly heterogeneous distributions
        if heterogeneity > thresh:
            # Calculate true variance of the Poisson Binomial
            true_variance = self.sum_p - self.sum_p_squared  # sum(p*(1-p))

            # Use refined parameter estimation
            binom_variance = self.n * p_avg * (1 - p_avg)

            if true_variance < binom_variance:
                # When true variance is lower, we need n_eff > n
                variance_ratio = binom_variance / true_variance if true_variance > 0 else 1.0
                n_eff = min(10 * self.n, self.n * variance_ratio)
            else:
                # When true variance is higher, we need n_eff < n
                variance_ratio = true_variance / binom_variance if binom_variance > 0 else 1.0
                n_eff = max(k, self.n / variance_ratio)  # Ensure n_eff >= k

            # Adjust p to maintain the same mean
            p_eff = self.sum_p / n_eff if n_eff > 0 else p_avg
            p_eff = max(min(0.9999, p_eff), 0.0001)  # Ensure it's a valid probability

            # Handle non-integer n_eff
            n_floor = np.floor(n_eff)
            n_ceil = np.ceil(n_eff)
            weight_ceil = n_eff - n_floor

            if n_floor == n_ceil:
                return binom.cdf(k, n_floor, p_eff)
            else:
                # Linear interpolation between binomials
                cdf_floor = binom.cdf(min(k, n_floor), n_floor, p_eff)
                cdf_ceil = binom.cdf(k, n_ceil, p_eff)
                return (1 - weight_ceil) * cdf_floor + weight_ceil * cdf_ceil
        else:
            # For nearly homogeneous probabilities, use standard binomial with
            # small adjustment based on cutoff point
            cutoff = self.n * p_avg  # Mean as the cutoff point

            if k > cutoff:
                # For high k values (right tail), slightly increase p
                adjustment = coeff1 * (p_avg - coeff3)
                p_adjusted = max(min(0.9999, p_avg + adjustment), 0.0001)
                return binom.cdf(k, self.n, p_adjusted)
            else:
                # For low k values (left tail), slightly decrease p
                adjustment = coeff2 * (p_avg - coeff3)
                p_adjusted = max(min(0.9999, p_avg - adjustment), 0.0001)
                return binom.cdf(k, self.n, p_adjusted)
```
Example:
```
    probabilities = np.asarray([0.9, 0.95, 0.9, 0.7, 0.8])
    n_succ = 1
    for p in probabilities:
        approximator.add_trial(p)
    pmf_1 = [approximator.pmf(n) for n in range(len(probabilities) + 1)]
    p_value_norm_approx = sum(pmf_1[0:n_success + 1])
```
They work quite differently (calculating the PMF vs CDF, returning p vs log(p), but hopefully it's still useful

L-M-Sherlock:


other.py
```diff
@@ -2934,6 +2935,106 @@ def create_features(df, model_name="FSRSv3"):
...
 
 
+class HighProbabilityBinomialApproximatorCDF:
+    def __init__(self):
+        self.sum_p = 0.0
+        self.sum_p_squared = 0.0
+        self.sum_log_q = 0.0  # For the boundary case
+        self.n = 0
+
+    def add_trial(self, p):
```
Expertium: We can remove add_trial and just make HighProbabilityBinomialApproximatorCDF a function instead of a class, I think it's more convenient.
sum_p, sum_p_squared, sum_log_q, n and k can be stored in the state. See the example at the end of the file: https://drive.google.com/file/d/1bku1aG_IJty0HhWATLvwwS6YRZHtHZ-m/view?usp=sharing

Expertium (changes requested):

other.py
Expertium: Actually, wait, why did you even implement the numpy version at all? My Google Drive file already contains a torch version, which is what we need

```python
def heterogeneous_binomial_log_cdf(sum_r, sum_r_squared, sum_log_1mr, n, k, device=None):
    """
    Calculate log probability that number of successes is <= k
    for a heterogeneous binomial distribution.

    Args:
        sum_r: Sum of probability values (tensor)
        sum_r_squared: Sum of squared probability values (tensor)
        sum_log_1mr: Sum of log(1-r) values (tensor)
        n: Number of trials (tensor)
        k: Success threshold (tensor)
        device: Computation device (default: None, will use available GPU or CPU)

    Returns:
        Log probability that number of successes <= k (tensor)
    """
    if device is None:
        device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')

    # Move inputs to device if they're not already there
    sum_r = sum_r.to(device)
    sum_r_squared = sum_r_squared.to(device)
    sum_log_1mr = sum_log_1mr.to(device)
    n = n.to(device)
    k = k.to(device)

    # Ensure integer type for k and n
    k = k.long()
    n = n.long()

    # Constants
    zero = torch.tensor(0.0, device=device)
    one = torch.tensor(1.0, device=device)
    tiny = torch.tensor(1e-10, device=device)
    small_val = torch.tensor(0.0001, device=device)
    large_val = torch.tensor(0.9999, device=device)
    large_neg = torch.tensor(-1e20, device=device)

    # Helper function for binomial log CDF using fully vectorized operations
    def binomial_log_cdf(k_input, n_input, p_input):
        """Calculate log of binomial CDF P(X ≤ k) where X ~ Bin(n, p)"""
        # Create tensor zeros of same shape as k_input for clamping
        zeros = torch.zeros_like(k_input)

        # Clamp k to valid range using tensor operations
        k_calc = torch.maximum(zeros, torch.minimum(k_input, n_input))

        # Create mask for valid calculations (n > 0)
        valid_mask = (n_input > 0)

        # For k >= n, result should be 0 (log(1))
        k_geq_n = (k_input >= n_input)

        # For remaining cases, calculate using vectorized operations
        # Create a mask for cases we need to calculate (n>0, k<n, k>=0)
        calc_mask = valid_mask & (~k_geq_n) & (k_calc >= 0)

        # Find max_k for our calculation range (across valid elements)
        max_k = torch.max(torch.where(calc_mask, k_calc, zeros))

        # Only proceed with the full calculation if there's at least one valid case
        has_valid = torch.any(calc_mask)

        # Create our result based on the calculation
        log_cdf_result = torch.zeros_like(k_input, dtype=torch.float32)

        # Use torch.where to conditionally compute the log CDF
        batch_size = k_input.shape[0]

        # Get j_max as integer (will be 0 if max_k < 0 or no valid cases)
        j_max = max_k.item()

        # Create a safety factor to handle j_max = 0 case
        j_range_size = torch.maximum(torch.tensor(1, device=device), max_k + 1)

        # Create range and expanded tensors (always create, even if j_max < 0)
        j_range = torch.arange(j_range_size, device=device)

        # Initialize tensors that will be used in the calculation
        j_mask = torch.zeros((j_range_size, batch_size), dtype=torch.bool, device=device)
        log_pmf = torch.zeros((j_range_size, batch_size), device=device)

        # Use torch.where with the condition (j_max >= 0 & has_valid) to conditionally execute the calculation
        calculation_needed = (max_k >= 0) & has_valid

        # This section is executed for all cases but only affects the result when calculation_needed
        if calculation_needed:
            # Expand dimensions for broadcasting
            j_range_expanded = j_range.unsqueeze(1).expand(-1, batch_size)
            n_expanded = n_input.unsqueeze(0).expand(j_range_size, -1)
            p_expanded = p_input.unsqueeze(0).expand(j_range_size, -1)
            k_expanded = k_calc.unsqueeze(0).expand(j_range_size, -1)

            # Mask for valid j values (j <= k) for each batch element
            j_mask = (j_range_expanded <= k_expanded) & calc_mask.unsqueeze(0)

            # Compute log PMF for all valid j values using the binomial formula
            log_n_choose_j = torch.lgamma(n_expanded + 1) - torch.lgamma(j_range_expanded + 1) - torch.lgamma(
                n_expanded - j_range_expanded + 1)

            log_pmf = log_n_choose_j + \
                      j_range_expanded * torch.log(p_expanded + tiny) + \
                      (n_expanded - j_range_expanded) * torch.log(1 - p_expanded + tiny)

            # Set invalid values to large negative number (effectively zero probability)
            log_pmf = torch.where(j_mask, log_pmf, large_neg)

            # Compute log-sum-exp for each batch element
            max_log_pmf_per_batch = torch.max(log_pmf, dim=0)[0]

            # Only apply logsumexp where we need to calculate
            log_sum_exp = max_log_pmf_per_batch + torch.log(
                torch.sum(torch.exp(log_pmf - max_log_pmf_per_batch.unsqueeze(0)), dim=0)
            )

            # Update result only for calc_mask elements
            log_cdf_result = torch.where(calc_mask, log_sum_exp, log_cdf_result)

        # Final result: use 0 (log(1)) when k>=n, calculated value otherwise
        final_result = torch.where(k_geq_n & valid_mask, zero, log_cdf_result)

        return final_result

    # Handle boundary cases
    k_eq_n = (k >= n) & (n > 0)
    k_eq_0 = (k == 0) & (n > 0)
    n_eq_0 = (n == 0)

    # For n=0, log(1)=0, since P(X ≤ k) = 1 when there are no trials
    boundary_result = torch.where(
        n_eq_0, zero,
        torch.where(k_eq_n, zero,
                    torch.where(k_eq_0, sum_log_1mr, zero))
    )

    # Only proceed with calculations if not a boundary case
    is_boundary = k_eq_n | k_eq_0 | n_eq_0

    # Get average probability and heterogeneity for all cases
    n_float = n.float()
    p_avg = torch.where(n_float > zero, sum_r / n_float, zero)

    # Calculate heterogeneity
    n_gt_1 = n_float > one
    p_variance = torch.where(n_gt_1,
                             (sum_r_squared / n_float) - (p_avg * p_avg),
                             zero)
    p_std_dev = torch.sqrt(torch.clamp(p_variance, min=0.0))
    cv = torch.where(p_avg > zero, p_std_dev / p_avg, zero)

    # Parameters
    coeff1 = one * 0.0256
    coeff2 = one * 0.0200
    coeff3 = one * 0.95
    thresh = one * 0.98

    # Calculate true variance
    true_variance = sum_r - sum_r_squared
    binom_variance = n_float * p_avg * (1 - p_avg)

    # Flag for heterogeneous case
    is_heterogeneous = cv > thresh

    # Heterogeneous case calculations
    true_var_lt_binom = true_variance < binom_variance

    # Calculate variance ratio safely
    var_ratio_low = torch.where(true_variance > zero,
                                binom_variance / true_variance, one)
    var_ratio_high = torch.where(binom_variance > zero,
                                 true_variance / binom_variance, one)

    # Apply right variance ratio based on comparison
    variance_ratio = torch.where(true_var_lt_binom, var_ratio_low, var_ratio_high)

    # Calculate n_eff based on variance comparison
    n_eff_low = torch.min(10 * n_float, n_float * variance_ratio)
    n_eff_high = torch.max(k.float(), n_float / variance_ratio)
    n_eff = torch.where(true_var_lt_binom, n_eff_low, n_eff_high)

    # Adjust p to maintain same mean
    p_eff = torch.where(n_eff > zero, sum_r / n_eff, p_avg)
    p_eff = torch.clamp(p_eff, min=small_val.item(), max=large_val.item())

    # Handle non-integer n_eff
    n_floor = torch.floor(n_eff).long()
    n_ceil = torch.ceil(n_eff).long()
    weight_ceil = n_eff - n_floor.float()

    # Calculate floor and ceil log CDFs
    k_floor = torch.minimum(k, n_floor)

    # Calculate CDFs for both floor and ceiling n values
    log_cdf_floor = binomial_log_cdf(k_floor, n_floor, p_eff)
    log_cdf_ceil = binomial_log_cdf(k, n_ceil, p_eff)

    # Handle equal floor/ceiling case
    floors_equal_ceils = n_floor == n_ceil

    # For interpolation
    log_weight_floor = torch.log(1 - weight_ceil + tiny)
    log_weight_ceil = torch.log(weight_ceil + tiny)

    max_log_cdf = torch.maximum(log_cdf_floor + log_weight_floor,
                                log_cdf_ceil + log_weight_ceil)

    weighted_log_cdf = max_log_cdf + torch.log(
        torch.exp(log_cdf_floor + log_weight_floor - max_log_cdf) +
        torch.exp(log_cdf_ceil + log_weight_ceil - max_log_cdf)
    )

    # Choose between floor-only or weighted result
    heterogeneous_result = torch.where(floors_equal_ceils, log_cdf_floor, weighted_log_cdf)

    # Homogeneous case calculations
    cutoff = n_float * p_avg
    is_right_tail = k > cutoff

    adjustment_right = coeff1 * (p_avg - coeff3)
    adjustment_left = coeff2 * (p_avg - coeff3)

    p_adjusted = torch.where(
        is_right_tail,
        torch.clamp(p_avg + adjustment_right, min=small_val.item(), max=large_val.item()),
        torch.clamp(p_avg - adjustment_left, min=small_val.item(), max=large_val.item())
    )

    homogeneous_result = binomial_log_cdf(k, n, p_adjusted)

    # Choose heterogeneous or homogeneous result based on the heterogeneity level
    calculation_result = torch.where(is_heterogeneous, heterogeneous_result, homogeneous_result)

    # Final result - use boundary if it's a boundary case, otherwise use calculation
    final_result = torch.where(is_boundary, boundary_result, calculation_result)

    return final_result
```

Usage example:
```python
sum_r = torch.tensor([0.0000, 1.8254, 1.8254, 1.8180, 1.7552,
        0.9270])
sum_r_squared = torch.tensor([0.0000, 1.6665, 1.6665, 1.6532, 1.5452,
        0.8594])
sum_log_1mr = torch.tensor([ 0.0000,  -4.9044, -4.9044, -4.8337,
        -4.3788, -2.6178])
n = torch.tensor([0., 2., 2., 2., 2., 1.])
k = torch.tensor([0., 0., 1., 2., 1., 1.])

print(heterogeneous_binomial_log_cdf(sum_r, sum_r_squared, sum_log_1mr, n, k))
```

I just need this function inside FSRS (replace `if calculation_needed:` with `torch.where` and clean up unused code).

Then we store the 5 summary statistics in the state (this piece is from `step`):
```python
        binary_grade = torch.where(X[:, 1] > 1, torch.ones_like(state[:, 0]), torch.zeros_like(state[:, 0]))
        sum_r = torch.where(short_term, state[:, 2], state[:, 2] + r)
        sum_r_squared = torch.where(short_term, state[:, 3], state[:, 3] + r * r)
        sum_log_1mr = torch.where(short_term, state[:, 4], state[:, 4] + torch.log(1 - r))
        n = torch.where(short_term, state[:, 5], state[:, 5] + 1)
        k = torch.where(short_term, state[:, 6], state[:, 6] + binary_grade)
    return torch.stack([new_s, new_d, sum_r, sum_r_squared, sum_log_1mr, n, k], dim=1)
```
Clamp R between 0.0001 and 0.9999 inside step, btw

Expertium:
So in case you are confused about what I want: https://github.com/open-spaced-repetition/srs-benchmark/pull/192#discussion_r2010918651

1) Implement a *function*, not a class
2) Remove unused code, replace if with torch.where
3) Add it to the FSRS-5 class or make a new FSRS-H class
4) Add summary statistics needed to calculate historical likelihood to the state (see that long comment)
5) Make sure it all works with the rest of the code

other.py
L-M-Sherlock: > Actually, wait, why did you even implement the numpy version at all?

I haven't checked the Google Drive. I just copy your pasted code.

L-M-Sherlock:
What's the unused code? And what is used to replace it? I don't understand the math behind the code, so I'm afraid to break something.

Edit: I try to refactor the code. Here is the result:

```
tensor([1.0000, 0.0074, 0.1656, 1.0000, 0.2273, 1.0000])
```

Expertium:
> What's the unused code? And what is used to replace it? I don't understand the math behind the code, so I'm afraid to break something.

image

This is the old Claude code, btw. I guess you already changed these parts.

> Edit: I try to refactor the code. Here is the result:
> 
> tensor([1.0000, 0.0074, 0.1656, 1.0000, 0.2273, 1.0000])

If this is based on these example inputs:
```
sum_r = torch.tensor([0.0000, 1.8254, 1.8254, 1.8180, 1.7552, 0.9270])
sum_r_squared = torch.tensor([0.0000, 1.6665, 1.6665, 1.6532, 1.5452, 0.8594])
sum_log_1mr = torch.tensor([ 0.0000,  -4.9044, -4.9044, -4.8337, -4.3788, -2.6178])
n = torch.tensor([0., 2., 2., 2., 2., 1.])
k = torch.tensor([0., 0., 1., 2., 1., 1.])
```
Then the correct output is `tensor([ 0.0000, -4.9044, -1.7981,  0.0000, -1.4816,  0.0000])`. I guess you forgot the logarithm. Speaking of which, make it return -log, so that the result is positive.

other.py
```diff
@@ -2935,104 +2935,109 @@ def create_features(df, model_name="FSRSv3"):
...
+        k: Number of successes to calculate CDF for (1-D tensor)
+
+    Returns:
+        Probability that number of successes is <= k (1-D tensor)
+    """
+    # Initialize result tensor
+    result = torch.zeros_like(k, dtype=torch.float32)
+
+    # Handle boundary cases
+    mask_neg = k < 0
```
Expertium: You can remove mask_neg, k will never be <0 in practice

L-M-Sherlock: Done

Expertium:
Alright, looks good, thank you. One last thing - can you make a new FSRS-H class with this function inside of it? And with this:
```python
def step(self, X: Tensor, state: Tensor) -> Tensor:
    """
    :param X: shape[batch_size, 2], X[:,0] is elapsed time, X[:,1] is rating
    :param state: shape[batch_size, 2], state[:,0] is stability, state[:,1] is difficulty
    # for calculating historical likelihood
    :param state[:,2] is sum_r, state[:,3] is sum_r_squared, state[:,4] is sum_log_1mr (log(1-R)),
    :param state[:,5] is n reviews (not same-day), state[:,6] is k successes (not same-day)
    :return state:
    """
    if torch.equal(state, torch.zeros_like(state)):
        ...
        # For historical likelihood
        sum_r = torch.zeros_like(state[:, 0], device=DEVICE)
        sum_r_squared = torch.zeros_like(state[:, 0], device=DEVICE)
        sum_log_1mr = torch.zeros_like(state[:, 0], device=DEVICE)
        n = torch.zeros_like(state[:, 0], device=DEVICE)
        k = torch.zeros_like(state[:, 0], device=DEVICE)
    else:
        r = self.forgetting_curve(X[:, 0], state[:, 0])
        r = r.clamp(0.0001, 0.9999)

        short_term = X[:, 0] < 1

        # For calculating historical likelihood
        binary_grade = torch.where(X[:, 1] > 1, torch.ones_like(state[:, 0]), torch.zeros_like(state[:, 0]))
        sum_r = torch.where(short_term, state[:, 2], state[:, 2] + r)
        sum_r_squared = torch.where(short_term, state[:, 3], state[:, 3] + r * r)
        sum_log_1mr = torch.where(short_term, state[:, 4], state[:, 4] + torch.log(1 - r))
        n = torch.where(short_term, state[:, 5], state[:, 5] + 1)
        k = torch.where(short_term, state[:, 6], state[:, 6] + binary_grade)

        success = X[:, 1] > 1
        new_s = torch.where(
            short_term,
            self.stability_short_term(state, X[:, 1]),
            torch.where(
                success,
                self.stability_after_success(state, r, X[:, 1]),
                self.stability_after_failure(state, r),
            ),
        )
        new_d = self.next_d(state, X[:, 1])
        new_d = new_d.clamp(1, 10)
    new_s = new_s.clamp(S_MIN, 36500)
    return torch.stack([new_s, new_d, sum_r, sum_r_squared, sum_log_1mr, n, k], dim=1)
```

L-M-Sherlock:
Which line do you want to place the `heterogeneous_binomial_log_cdf` function?

Expertium:
Uh, idk. Anywhere, man.

L-M-Sherlock:
OK, please check the latest commit.

other.py
```diff
@@ -1145,6 +1146,237 @@ def state_dict(self):
...
+        """
+        :param X: shape[batch_size, 2], X[:,0] is elapsed time, X[:,1] is rating
+        :param state: shape[batch_size, 2], state[:,0] is stability, state[:,1] is difficulty
+        # for calculating historical likelihood
+        :param state[:,2] is sum_r, state[:,3] is sum_r_squared, state[:,4] is sum_log_1mr (log(1-R)),
+        :param state[:,5] is n reviews (not same-day), state[:,6] is k successes (not same-day)
+        :return state:
+        """
+        if torch.equal(state, torch.zeros_like(state)):
+            ...
```
Expertium, line 1309: ... was meant to be a placeholder, there should be code related to S0 and D0

Expertium (changes requested):

other.py
L-M-Sherlock: We need to distinguish learning and review via it. What's the initial state for your design?

Expertium: They all are initialized to 0. As for S0 and D0, just leave them as in FSRS-5
```
        sum_r = torch.zeros_like(state[:, 0], device=DEVICE)
        sum_r_squared = torch.zeros_like(state[:, 0], device=DEVICE)
        sum_log_1mr = torch.zeros_like(state[:, 0], device=DEVICE)
        n = torch.zeros_like(state[:, 0], device=DEVICE)
        k = torch.zeros_like(state[:, 0], device=DEVICE)
```
Btw, since the length of the memory state changed, shouldn't there be changes in some other parts of other.py?

L-M-Sherlock: Please see the `forward` function. The `state` is initialized there.

other.py
```diff
@@ -1144,6 +1146,237 @@ def state_dict(self):
...
+        )
+
+    def forward(
+        self, inputs: Tensor, state: Optional[Tensor] = None
+    ) -> tuple[Tensor, Tensor]:
+        """
+        :param inputs: shape[seq_len, batch_size, 2]
+        """
+        if state is None:
+            state = torch.zeros((inputs.shape[1], 2))
```
L-M-Sherlock: @Expertium, if I understand correctly, the `2` should be replaced with `7`?

Expertium: I think so

Expertium:
Tell me when everything is ready to be used. Then I'll download the code from this branch and try incorporating historical likelihood into FSRS formulas and report the results

Expertium:
@L-M-Sherlock @1DWalker 

I tested using H like this:
```
            w[19] = w[19].clamp(0.01, 1)
            w[20] = w[20].clamp(0.01, 1)
            
    def stability_after_success(self, state: Tensor, r: Tensor, h: Tensor, rating: Tensor) -> Tensor:
        hard_penalty = torch.where(rating == 2, self.w[15], 1)
        easy_bonus = torch.where(rating == 4, self.w[16], 1)
        new_s = state[:, 0] * (
            1
            + torch.exp(self.w[8])
            * (11 - state[:, 1])
            * torch.pow(1 + h, -self.w[19])
            # * torch.exp(-self.w[19] * h)
            * torch.pow(state[:, 0], -self.w[9])
            * (torch.exp((1 - r) * self.w[10]) - 1)
            * hard_penalty
            * easy_bonus
        )
        return new_s

    def stability_after_failure(self, state: Tensor, r: Tensor, h: Tensor) -> Tensor:
        old_s = state[:, 0]
        new_s = (
            self.w[11]
            * torch.pow(state[:, 1], -self.w[12])
            # * torch.pow(1 + h, -self.w[20])
            * torch.exp(-self.w[20] * h)
            * (torch.pow(old_s + 1, self.w[13]) - 1)
            * torch.exp((1 - r) * self.w[14])
        )
        new_minimum_s = old_s / torch.exp(self.w[17] * self.w[18])
        return torch.minimum(new_s, new_minimum_s)
```
The difference in performance was not statistically significant. Not only that, but it made the optimizer 1.7-2 times slower.
image

---

Unrelated, but I tested a different idea: making the hard_penalty go below 0.
The naive way to do it is like this:
`            w[15] = w[15].clamp(-1, 1)  # Hard penalty`
That's it.
However, this creates 2 problems:
1) New S when the user presses Hard could end up being **lower** than post-lapse S aka PLS. Not only that, but it could end up being **negative**. Neither makes sense.
2) Cards with lower difficulty would have **lower** new S and cards with higher difficulty would have **higher** new S, which makes no sense.

So instead I implemented this:
```
            w[15] = w[15].clamp(-1, 1)  # Hard penalty
            
                def stability_after_success(self, state: Tensor, r: Tensor, rating: Tensor) -> Tensor:
        hard_penalty = torch.where(rating == 2, self.w[15], 1)
        pls = self.stability_after_failure(state, r)
        # if hard_penalty <= 0, then SInc = 1 and final S = PLS * (1 + hard_penalty) + prev_S * (1 - (1 + hard_penalty))
        # = PLS * (1 + hard_penalty) + prev_S * -hard_penalty
        # In other words, if hard_penalty <= 0, we take a weighted average of PLS and previous S, otherwise we use new S
        # if hard penalty > 0, then SInc > 1 and final S = new_S
        # if hard penalty = -1, then final S = PLS
        pls_weight = torch.where(hard_penalty <= 0, 1 + hard_penalty, 0)  # the weight of PLS is non-zero only when hard_penalty < 0
        easy_bonus = torch.where(rating == 4, self.w[16], 1)
        new_s = state[:, 0] * (
            1
            + torch.exp(self.w[8])
            * (11 - state[:, 1])
            * torch.pow(state[:, 0], -self.w[9])
            * (torch.exp((1 - r) * self.w[10]) - 1)
            * torch.maximum(hard_penalty, torch.zeros_like(rating))  # Ensure SInc itself is never < 1
            * easy_bonus
        )
        # return new_s
        result = pls * pls_weight + new_s * (1 - pls_weight)
        return result
```
It's probably not immediately clear what is going on, so let me explain.

### w[15] now operates in two distinct regimes: when it's >0, it acts as a multiplier in the SInc formula. When it's <0, it acts as a weight in a weighted average of the previous S and PLS, SInc=1 in this case.

In practice it means that:
1) If w[15]=-1, new S(Hard) = new S(Again)
2) If w[15]=0, new S(Hard) = previous S
3) If w[15]>0, new S(Hard) = SInc(grade, D, S, R)
4) If w[15]=1, new S(Hard) = new S(Good)

This solves both the problem of new S being less than PLS or even less than 0 by tying new S to PLS, and the problem of S of easy cards "resetting" harder than S of difficult cards. Now this satisfies all intuitively desirable properties while adding more flexibility. So, does it work? Well...
image
..yesn't. I double- and tripple-checked that the Wilcoxon test isn't bugged. Nope, it works just fine.

---

Unrelated again, but I asked Claude 3.7 to replace next_d with a neural net to see if it could improve the accuracy.
Relevant code:
```
    def __init__(self, w: List[float] = init_w):
        super(FSRS5, self).__init__()
        self.w = nn.Parameter(torch.tensor(w, dtype=torch.float32))
        self.init_w_tensor = self.w.data.clone().to(DEVICE)
        self.dropout = 0.0

        # Neural network for next_d function
        self.difficulty_nn = nn.Sequential(
            nn.Linear(3, 16),  # Input: retrievability, difficulty, rating
            nn.Dropout(self.dropout),
            Mish(),
            nn.Linear(16, 16),
            nn.Dropout(self.dropout),
            Mish(),
            nn.Linear(16, 1)  # Output: new difficulty
        )

        # Initialize weights using Xavier/Glorot initialization
        for m in self.difficulty_nn.modules():
            if isinstance(m, nn.Linear):
                nn.init.xavier_normal_(m.weight)
                nn.init.constant_(m.bias, 0)

    # Custom Mish activation function
    class Mish(nn.Module):
        def forward(self, x):
            # Numerically stable version
            delta = torch.log1p(torch.exp(x))
            # For large x values, use an approximation
            mask = x > 20  # Threshold where exp(x) would cause an overflow
            delta = torch.where(mask, x, delta)
            return x * torch.tanh(delta)
            
    def next_d(self, r: Tensor, last_d: Tensor, rating: Tensor) -> Tensor:
        nn_input = torch.cat([
            r.unsqueeze(1),
            last_d.unsqueeze(1),
            rating.unsqueeze(1)
        ], dim=1)

        new_d_raw = self.difficulty_nn(nn_input).squeeze(1)

        init_d = self.init_d(4)
        new_d = self.mean_reversion(init_d, new_d_raw)

        return new_d
```
Here are the results
image
What, did anyone actually think I can implement something like this on my own? I freakin' wish. This is why I keep pestering Alex and Jarrett to do this.

EDIT: actually, no, it works if I put the Mish class outside of the FSRS class. Because of course. Because it makes sense. Because reasons, yeah, of course.
Alright, I'll try it with different dropout rates and report the results later (actually, I should use weight decay or something instead of dropout). I feel like it won't work well without pretraining it on 100 users first.

Expertium:
I couldn't get the neural D to work, so I'll be waiting for Jarrett to do it.
@L-M-Sherlock you can close this PR since historical likelihood didn't help to improve FSRS

L-M-Sherlock:
> I couldn't get the neural D to work, so I'll be waiting for Jarrett to do it.

Please open a new issue to discuss it. I will take a look later.

PR #200, opened 2025-04-07
Added more info about the two- and three-component model of memory

Expertium:
It's also clearer now; previously, the text was implying that Ebisu and HLR use a three-component model, which is not true.


L-M-Sherlock (approved):
LGTM

PR #201, opened 2025-04-11
Expt/trainable forgetting decay

L-M-Sherlock:


README.md
```diff
@@ -39,14 +39,15 @@ Log Loss and RMSE (bins) measure calibration: how well predicted probabilities o
...
-        - FSRS-5 pretrain: FSRS-5 where only the first 4 parameters (values of initial stability after the first review) are optimized and the rest are set to default.
-        - FSRS-5 binary: FSRS which treats `hard` and `easy` grades as `good`.
-        - FSRS-5 preset: different parameters are used for each preset. The minimum number of presets in Anki is one, a preset can be applied to multiple decks.
-        - FSRS-5 deck: different parameters are used for each deck.
-        - FSRS-5 recency: FSRS-5 trained with reviews being weighted based on their recency, such that older reviews affect the loss function less and newer reviews affect it more.
-    - FSRS-rs: the Rust port of FSRS-5. See also: https://github.com/open-spaced-repetition/fsrs-rs
+    - FSRS-5: the upgraded version of FSRS. Unlike the previous versions, it uses the same-day review data. Same-day reviews are used only for training, and not for evaluation.
+    - FSRS-6: the latest version of FSRS. It fixes some issues of the same-day reviews and makes the decay of the forgetting curve trainable.
```
Expertium: Suggested change:
"The formula for handling same-day reviews has been improved. More importantly, FSRS-6 has an optimizable parameter that controls the flatness of the forgetting curve, meaning that the shape of the curve is different for different users."

Expertium (changes requested):
For the sake of not making the table cluttered, please add -deck, -recency, -pretrain, -binary and other variants ONLY for the most recent version of FSRS. Remove FSRS-5-recency and all other variants from the table, there should only be one variant for pre-FSRS-6 versions.

Expertium (changes requested):
For the sake of not making the table cluttered, please add -deck, -recency, -pretrain and other variants ONLY for the most recent version of FSRS. Remove FSRS-5-recency and all other variants, there should only be one variant for pre-FSRS-6 versions.

L-M-Sherlock:
OK. I haven't completed the remaining benchmarks for FSRS-6 variants. It will take 2 or 3 days.

Expertium:
Speaking of which, I think we can improve this further.

We are esestimating S0 with a fixed decay, but why not make decay trainable here as well?

Make `minimize` estimate both S0 and decay. The problem is that the values of decay for different grades will be different. Still, if we take a weighted (by n reviews for each grade) average and use that as the starting value of decay for further optimization via gradient descent, it will likely be better.

1) Modify the pretrain process to obtain 4 values of S0 and 4 values of decay.
2) Calculate weighted average decay.
3) Pass 4 values of S0 and one the value of weighted average decay into `init_w`.

L-M-Sherlock:
> We are esestimating S0 with a fixed decay, but why not make decay trainable here as well?

Because I cannot do the same thing in Rust, unless I refactor the current code.

The current method is based on ternary search:

https://github.com/open-spaced-repetition/fsrs-rs/blob/1c2bd3d9fe83592a49056b14394b4ed299526777/src/pre_training.rs#L120-L165

Expertium:
Well, time to find some Rust library for numerical minimization, lol
Or ask Claude 3.7 to write you some code for that

Here: https://lib.rs/crates/argmin

L-M-Sherlock:
😅fine, I would re-run the benchmarking again. Maybe I need more time to make sure the consistency between the Python and Rust implementations.

Expertium:
Btw, Claude recommends using either the particle swarm method or the Nelder-Mead method for optimizing two (S0, decay) bounded parameters (https://docs.rs/argmin/latest/argmin/)

L-M-Sherlock:
> Btw, Claude recommends using either the particle swarm method or the Nelder-Mead method for optimizing two (S0, decay) bounded parameters (https://docs.rs/argmin/latest/argmin/)

I have let Claude implement it. But the Nelder-Mead method cannot find the optimum.

Edit: ...I even cannot set boundaries for the parameters.

Expertium:
You don't need to implement it though, that's what the library is for. Unless you mean "I let Claude write the code using said library".
Are you saying that this library doesn't allow you to set boundaries for parameters?

L-M-Sherlock:
> Are you saying that this library doesn't allow you to set boundaries for parameters?

Yep. The library doesn't provide this option.

Expertium:
> GRU: a type of recurrent neural network that's often used for making predictions based on a sequence of data. It's a classic in the field of machine learning for time-related tasks. It uses the same power forgetting curve as FSRS-4.5 and FSRS-5 to make the comparison more fair.

Should GRU have an optimizable curve now?

L-M-Sherlock:
We have GRU-P.

PR #202, opened 2025-04-11
Weight -> Parameter value

Expertium:


Expertium:
Also, consider changing x limits for w[0], w[1] and w[2], so that most of the space on the graph isn't empty
image
This just looks unreadable

L-M-Sherlock (approved):

PR #205, opened 2025-04-22
Fix/Ignore extra return from FSRSv1

Luc-Mcgrady:
Running
```
python other.py --algo FSRSv1
```
raises an error.


L-M-Sherlock (approved):
LGTM

PR #206, opened 2025-04-25
Updated outdated info in Superiority

Expertium:
Also, while I doubt that anyone cares about the full table, I still recommend fixing this issue with overlapping text:
image


L-M-Sherlock (approved):
LGTM

PR #207, opened 2025-04-26
Alex's uncheatable RMSE

Expertium:
@1DWalker I brought you some delicious AI slop :D

Check if this is indeed what you have been talking about or if I got it wrong

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -17,39 +17,178 @@ def catch_exceptions(func):
 
 
 def rmse_matrix(df):
+    """
+    Args:
+        df (pd.DataFrame): DataFrame containing review data with columns like
+                           'r_history', 't_history', 'elapsed_days', 'i',
+                           'y' (true outcome, e.g., 0 or 1), 'p' (predicted recall prob),
+                           and optionally 'weights'.
+    Returns:
+        float: The calculated root mean squared error. Returns np.nan if
+               calculation is not possible (e.g., empty data after processing).
+    """
+    if df.empty:
+        print("Warning: Input DataFrame is empty.")
+        return np.nan
+
+    # Ensure required columns exist (adjust based on actual minimum requirements)
+    required_cols = ['r_history', 't_history', 'elapsed_days', 'i', 'y', 'p']
+    if not all(col in df.columns for col in required_cols):
+        missing_cols = [col for col in required_cols if col not in df.columns]
+        raise ValueError(f"Missing required columns: {missing_cols}")
+
     tmp = df.copy()
 
+    # --- Helper function to count lapses ---
     def count_lapse(r_history, t_history):
         lapse = 0
-        for r, t in zip(r_history.split(","), t_history.split(",")):
-            if t != "0" and r == "1":
-                lapse += 1
+        # Defensive coding: ensure inputs are strings and handle potential errors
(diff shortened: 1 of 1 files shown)


1DWalker:
I think it should only take a couple lines of code.

L-M-Sherlock:
If I understand it correctly, these changes would work:

```diff
diff --git a/utils.py b/utils.py
index 01bb1fd03c4..b0448661ccf 100644
--- a/utils.py
+++ b/utils.py
@@ -44,11 +44,9 @@ def rmse_matrix(df):
     )
     if "weights" not in tmp.columns:
         tmp["weights"] = 1
-    tmp = (
-        tmp.groupby(["delta_t", "i", "lapse"])
-        .agg({"y": "mean", "p": "mean", "weights": "sum"})
-        .reset_index()
-    )
+
+    tmp["y"] = tmp.groupby(["delta_t", "i", "lapse"])["y"].transform("mean")
+
     return root_mean_squared_error(tmp["y"], tmp["p"], sample_weight=tmp["weights"])
```

L-M-Sherlock:
Btw, is it still cheatable if the model always predicts the average R for each group?

1DWalker:
It is still cheatable, just less so. If there is only one global bin then this would be cheatable by predicting average R of the collection, but on the other hand if you define each bin to correspond with exact card histories (down to nanosecond resolution), then there is no more cheating, the best strategy aligns with that of log loss since this is just normal RMSE. The problem in this case is that we only get one 0/1 as the observed output and we lose the Anki manual interpretation of "difference between the predicted probability of recalling a card (R) and the actual probability". We need parallel universes to truly measure this actual probability. 
Perhaps we can create bins that are reasonable in resolution but also include enough cards to get a good predicted R for that bin. I'm not sure if it's possible and the current rounding scheme might exclude too much. I've worked on a toy model that takes each review, finds its bin, and predicts the average so far for that bin. From what I remember it wasn't much better than AVG on log loss. To be honest I'm kind of doubtful that this idea will work out but it is still better than the current RMSE (bins) which is too easily cheated.

L-M-Sherlock:
But I'm afraid that this method has another issue: a perfect prediction will be punished. For example, what if two kind of reviews fall into the same bins? A perfect model will output two level predicted probability and get a non-zero RMSE(bins) because they are different from the average probability.

1DWalker:
Yeah, optimality is not the same as with log loss unless bins = histories. But optimality doesn't align for the current RMSE (bins) either. We are making tradeoffs with any RMSE idea that isn't just the normal RMSE, the question is where do we want to make the tradeoffs.

L-M-Sherlock:
@Expertium thoughts? Changing the definition of RMSE(bins) requires to re-benchmark all algorithms, so we need consider it cautiously.

Expertium:
I'm not sure. I like the idea of uncheatable (or at least less cheatable) RMSE, but re-benchmarking everything is a pain. I'll defer to Alex.

Expertium:
@1DWalker @L-M-Sherlock come on guys, let's decide

1DWalker:
Let's keep the current version. It works well as a calibration metric and it is no longer optimized for in Anki so it is no longer harmful.

PR #209, opened 2025-04-27
Remove RMSE (not binned)

Expertium:
It serves zero purpose


L-M-Sherlock:
It doesn't take too much disk, so just keep it as it is. Or wait for the next re-benchmark.

Expertium:
@L-M-Sherlock do you plan to remove it now? If not, close this

PR #212, opened 2025-05-10
Modify rounding

Expertium:
See https://github.com/open-spaced-repetition/fsrs-vs-sm17/pull/72

### DIFF
--- a/evaluate.py
+++ b/evaluate.py
@@ -14,13 +14,13 @@ def sigdig(value, CI):
     CI_sigdigs = 2
     decimals = n_lead_zeros_CI + CI_sigdigs
     rounded_CI = round(CI, decimals)
-    rounded_value = round(value, decimals - 1)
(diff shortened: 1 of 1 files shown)


Expertium:
Basically, we're just reporting one more digit of the mean than with my previous function, that's it

Expertium:
Btw, I think this change is also good because some people might wonder "Huh, why is the number of digits after the decimal different for the mean and the CI?"
For example, 0.**312**±0.**0078**
That's how it is before my new change
With the new change it will be 0.**3120**±0.**0078**, for example, which won't make people question why the number of digits after the decimal point is different

L-M-Sherlock (approved):

Expertium:
@L-M-Sherlock don't forget to re-run evaluate.py and update the table. Same goes for the SM-17 repo

PR #215, opened 2025-05-16
Remove NN-17

Expertium:
I want to remove NN-17 since it's no longer clear whether its performance is in any way close to SM-17, or how well SM-17 itself performs, for that matter.


Expertium:
You'll have to re-make superiority tables and statistical significance tables

PR #216, opened 2025-05-17
Superiority -> Outperformance rate

Expertium:
I feel like the connotations of "superiority" in English aren't quite right, and it's better if we change it to something more neutral, like "outperformance rate". Feel free to suggest other names, like "comparative win rate" or something, but it needs to be short and easy to remember.
@1DWalker @user1823 @Luc-Mcgrady

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,9 +176,9 @@ Averages weighted by the number of reviews are more representative of "best case
 
 Unweighted averages are more representative of "average case" performance. In reality, not every user will have hundreds of thousands of reviews, so the algorithm won't always be able to reach its full potential.
 
-### Superiority
+### Outperformance Rate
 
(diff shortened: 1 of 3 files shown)


Expertium:
Well, I pinged 3 people and asked another 2, and nobody responded. It's up to you, Jarrett

Luc-Mcgrady:
Theres a cost that comes with people having to adjust to anything being renamed (think weights and parameters). I'm not sure why "superiority" should be ridded of its positive connotation when an algorithm which outperforms another algorithm more often should be concidered as better(?). "outperform" has this same connotation anyway (allbeit to a lesser extent). No strong feelings on this though.

Expertium:
@L-M-Sherlock

L-M-Sherlock:
I don't recommend updating the filenames.

Expertium:
Why not?

Luc-Mcgrady:
Are we going through with this when 3/5 people you pinged disagree with it?

Expertium:
I guess no

PR #218, opened 2025-05-22
Adds the RWKV model

1DWalker:
The code that supports fast parallelism is complicated, so I've included an rnn version that shows the information flow more clearly.


L-M-Sherlock:
Awesome! I will try to run it in this weekend. 

Edit: would you minding writing an article to illustrate how RWKV predicts the retrievability? And how to use RWKV to schedule the cards?

rwkv/README.md
```diff
@@ -0,0 +1,39 @@
...
+```bash
+pip install --no-build-isolation -e .
+```
+
+Create a helper db, used to precompute bins for the RMSE (bins) metric and to find which reviews are used in the benchmark. This requires 7 GB of storage.
+```bash
+python -m rwkv.find_equalize_test_reviews --config rwkv/find_equalize_test_reviews_config.toml
+```
+
+Preprocess the 10k dataset. With the default configuration, this will require ~400 GB of available storage.
```
L-M-Sherlock, line 17: OMG💀

I need a new SDD for that.

1DWalker:
To be honest, there isn't much to explain for RWKV, it just boils down to that I pushed the dataset through a neural network and to no one's surprise it does pretty well. The main difference between this and models like FSRS, LSTM, is that it is a bigger model and it uses all reviews seen so far when formulating its predictions. FSRS/LSTM can technically also incorporate all reviews so far by just doing more optimization after each review, but this is a bit too slow and so they rely on a 5-way split.

An analogy for RWKV-P is that similarly to an LLM, it does next-word predictions. But instead the sentences look like
"Query_card_1, pass. Query_card_2 fail. Query_card_1 fail". Predicting the next word, "pass" or "fail", corresponds to predicting R for a review. With this formulation we can benefit from LLM architectures, and [RWKV](https://github.com/BlinkDL/RWKV-LM) is one of them.

For scheduling, RWKV could work like how FSRS works right now for scheduling by targeting a threshold R. RWKV-P is not as reliable for scheduling, the simplest method would be to query random cards for their current estimated R and schedule them with the same threshold strategy, but this method has many pitfalls.

rwkv/README.md
1DWalker: You can reduce the number of users that are included and train a smaller version of the network if you want. The training code and configuration are not exactly what I used, I pretty much ran the training code until I ran out of patience, around 2.5 days in for each half of RWKV (one was trained on one half of the dataset, the other on the other half), I may have also randomly tinkered with things like regularization over time, so it isn't exactly reproducible.

rwkv/model/srs_model_rnn.py
```diff
@@ -0,0 +1,267 @@
...
+
+        self.w_linear = torch.nn.Linear(self.w_head_dim, self.num_curves)
+
+        self.s_point_spread = 18.5
+        self.s_max = 22
+
+        self.p_linear = torch.nn.Linear(self.p_head_dim, 4)
+
+    def forgetting_curve(self, w, label_elapsed_seconds):
+        s_space_raw = torch.exp(
```
L-M-Sherlock, line 76: If I understand it correctly, `w` will change every time the user does a review on any card? If so, does it mean RWKV will change all existing cards' memory state after each review?

1DWalker: `w` can be treated as `S` in FSRS, in the same way that `S` is specific to a review, `w` is also specific to a review. `w` is a distribution over exponential curves.

1DWalker: Also, `def interp(self, out_ahead_logits, label_elapsed_seconds)` allows RWKV to correct the raw curve by using local adjustments, this makes the curve not necessarily monotonic. But the performance gain is so little that I don't think it's important to worry about.

L-M-Sherlock: But `S` is standalone for each card. `w` is a global state if I understand it correctly, because it considers all historical reviews from all cards.

1DWalker: I think I understand the misconception. `w` is not the global state, but a property of specific reviews, like how you cannot say that the current S for a card was used for all reviews for that card. The RWKV state is defined [here](https://github.com/1DWalker/srs-benchmark/blob/9106bd713ecc36a6ce8375fa8c9122202389dfbc/rwkv/run_as_rnn.py#L53-L57).

L-M-Sherlock:
OK, I ran the training script successfully on my device after decreasing the `LMDB_SIZE`.

PR #221, opened 2025-05-30
set_num_threads to 1

Expertium:
image
image
Based on these two screenshots you would not be able to guess which one was done with `torch.set_num_threads(1)` and which one with `torch.set_num_threads(2)`. Let's set it to 1 then. It's better to use `torch.set_num_threads(1)` and use more `--processes`


L-M-Sherlock (approved):
LGTM

PR #223, opened 2025-06-04
re-benchmark FSRS-6 with new default parameters

L-M-Sherlock:


Expertium:
I don't see an update of readme

PR #230, opened 2025-06-11
Fix/comparable train_equals_test

### LINKED ISSUE #224: FSRS-6-recency-train_equals_test is not evaluated on the correct reviews
In the result file, the first user has `size: 12746` while all other algorithms have `size: 10620`.

L-M-Sherlock:
fix #224


1DWalker:
LGTM

PR #251, opened 2025-07-21
RWKV setup changes

RlzHi:
A change to allow pip to build successfully, and an optimisation to allow 100% CPU without freezing the OS. Please do review


L-M-Sherlock (approved):
LGTM

PR #252, opened 2025-07-21
Additional plots

RlzHi:
Additional plots from a request in discord, if you would like to include them.

image

Edit: [Has since been updated](https://github.com/open-spaced-repetition/srs-benchmark/pull/252#issuecomment-3102928220)


L-M-Sherlock:
I see `largest_average` and `all_average` in your notebook, and the plots are based on `all_average`. Is it intended?

RlzHi:
@L-M-Sherlock Great spot, was using that for some other analysis before the refactor. Should be as expected now.

L-M-Sherlock:
image

Does it means there is a user who did ~9460 reviews per day over 90 days?

RlzHi:
Yes, that's correct. There was someone in discord who did 10k/day over the past month too, so these users do exist. I do wonder how it's possible

L-M-Sherlock:
Could you also add a plots for the review durations? I wonder whether it would exceed 86400 seconds.😂

Expertium:
I think it makes sense to also add a graph that uses an average across the entire review history of that user

RlzHi:
@Expertium 
<details><summary>Graph now includes the full review history</summary>

(Edited to show graph after commit 642c493)
image
</details>

@L-M-Sherlock 
<details><summary>Added a notebook for the review durations - We have a review that's over 40,000 seconds (11 days)!</summary>

image
image
image
</details>

L-M-Sherlock (approved):
LGTM, thanks~

PR #255, opened 2025-07-29
Changed s_min_base

Expertium:
I think 1e-6 is too low, I plan to use 0.0001


L-M-Sherlock:
... I have to re-run the benchmark again.

PR #257, opened 2025-08-02
FSRS-7

Expertium:


Expertium:
For some reason I need to use both `--short` and `--secs`, unlike with FSRS-6. Idk why

Expertium:
I'll modify `init_w`, `default_params_stddev_tensor` and the constants in pretrain later. Oh, and we'll need to do the big pretrain on 1000 users, so we'll need to call those guys with a big cluster again

Expertium:
Idk why
image

models/fsrs_v7.py
```diff
@@ -0,0 +1,450 @@
...
+            # w[16] = w[16].clamp(1.5228, 3.0571)
+            # # Forgetting curve
+            # w[17] = w[17].clamp(0.0100, 0.0560)
+            # w[18] = w[18].clamp(0.1000, 0.2267)
+            # w[19] = w[19].clamp(0.1001, 0.4000)
+            # w[20] = w[20].clamp(0.3648, 0.4650)
+            module.w.data = w
+
+
+class FSRS7(FSRS6):
```
Luc-Mcgrady, line 76: https://github.com/open-spaced-repetition/srs-benchmark/actions/runs/16691373355/job/47249561800?pr=257#step:6:781

It's because you changed the parameters of `stability_after_success` so it doesn't inherit properly.

Expertium: But `stability_after_success` in FSRS-7 is the same as in FSRS v4. I might still add extra params though
image
image

models/fsrs_v7.py
```diff
@@ -0,0 +1,450 @@
...
+    ):
+        factor1 = 0.9 ** (1 / decay1) - 1
+        factor2 = 0.9 ** (1 / decay2) - 1
+        t_over_s = t / s
+        R1 = (1 + factor1 * t_over_s) ** decay1
+        R2 = (1 + factor2 * t_over_s) ** decay2
+        blending_function = blend_start * 2.718281828459045 ** (-k * t_over_s)
+        return blending_function * R1 + (1 - blending_function) * R2
+
+    def stability_after_success(
```
Luc-Mcgrady:
```suggestion
    def stability_after_success(  # type: ignore[override]
```

Ok XD

Luc-Mcgrady: Mark the above one as resolved as well I can't do it.

Expertium:
I couldn't get checks to pass with s_min=0.001, so I reverted it back to 0.0001. Well, I'll probably end up using 0.0001 anyway, it seems like 0.001 makes metrics worse

Expertium:
@Luc-Mcgrady I don't understand this stuff. I was trying to add https://github.com/open-spaced-repetition/srs-benchmark/commit/033cdaa4ef4f24057a8742580f2d4e5b6101265a to this branch, but it causes a conflict
EDIT: alright, I think it works now

Expertium:
I'll close this and make a new PR

PR #259, opened 2025-08-08
Feat/Use file path instead of filters

Luc-Mcgrady:
Apparently mildly faster

Figure_1

A = Filter
B = This pr

Tested with: https://github.com/Luc-Mcgrady/srs-benchmark/tree/load-test

I have only tested this with `script.py`


L-M-Sherlock (approved):
LGTM

PR #262, opened 2025-09-14
CONST that predicts a fixed constant p(recall)

Expertium:
Btw, we should remove "90%" since this is just a more general version of that.

### DIFF
--- a/config.py
+++ b/config.py
@@ -33,6 +33,7 @@ ModelName = Literal[
     "DASH[ACT-R]",
     # Other models
     "AVG",
+    "CONST",
     "RMSE-BINS-EXPLOIT",
     "MOVING-AVG",
     "90%",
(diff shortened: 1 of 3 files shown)


Expertium:
I don't think it needs feature engineering, but I can't figure out how to make it work otherwise

Expertium:
This is with 99.99%, `python other.py --algo CONST`

```
Model: CONST
Total number of users: 9999
Total number of reviews: 349923850
Weighted average by users:
CONST LogLoss (mean±std): 1.4311±1.1404
CONST RMSE(bins) (mean±std): 0.1837±0.1307

Weighted average by reviews:
CONST LogLoss (mean±std): 1.2341±0.9252
CONST RMSE(bins) (mean±std): 0.1580±0.1071
```

We can have 3 levels of baselines
CONST (always 99.99%) is a "if you can't beat this, go home" baseline
AVG is a "if you can't beat this, improve your algorithm" baseline
MOVING-AVG is a "if you CAN beat this, contact us, we need you on the team" baseline

1DWalker:
Why not const as 90%? It's more natural since the default DR is 90%

Expertium:
I just wanted to try one that predicts almost 100% aka "the card is always recalled"

1DWalker:
Then what about 99.9999? It gets pretty arbitrary due to the log loss calculation

PR #266, opened 2025-09-28
Remove requirements.txt

joshdavham:
I was initially confused when getting started with this repo as there's both a `requirements.txt` and a `pyproject.toml`. After analyzing both files, I noticed that they're redundant so I think it's a good idea to remove `requirements.txt`.

Lmk if you have any questions 👍

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,0 @@
-pandas
-matplotlib
-scikit-learn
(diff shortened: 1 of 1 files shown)


L-M-Sherlock (approved):
I forgot to remove it 😂

PR #281, opened 2025-11-30
Feat/Mean Bias Error

L-M-Sherlock:
| Algorithm | MBE |
|-|-|
|FSRS-6|0.23%|
|FSRS-5|-0.35%|
|FSRS-4.5|-0.47%|
|DASH|1.06%|
|FSRS-6 default|0.22%|
|Anki default|-7.4%|
|SM2|-8.74%|

Mean Bias Error could tell us whether an algorithm tends to underestimate the recall probability or overestimate it.

Underestimation usually means that the algorithm will give a shorter interval for the given desired retention. Overestimation means the algorithm tend to give a longer interval than the ideal.


Expertium:
When adding it to readme, please specify whether a positive value means overestimation or underestimation.
Btw, the value for FSRS-6 default is missing the % symbol in your table

L-M-Sherlock:
I don't have the plan to add them to readme. I add this code just out of curiosity.

PR #288, opened 2026-01-15
Enable LSTM duration only via --duration

L-M-Sherlock:


L-M-Sherlock:
@gemini-code-assist what's the duplicated logic for filename generation can be simplified by using the existing method in the Config class in reptile_trainer.py?

PR #290, opened 2026-01-23
FSRS-7

Expertium:
I have been doing all tests with an outdated version of the benchmark code, so I will run this with the updated version later. For now this will be a draft.

### DIFF
--- a/config.py
+++ b/config.py
@@ -13,6 +13,7 @@ ModelName = Literal[
     "FSRS-5",
     "FSRS-6",
     "FSRS-6-one-step",
+    "FSRS-7",
     "FSRS-rs",
     # Neural networks
     "RNN",
--- a/features/factory.py
+++ b/features/factory.py
@@ -28,6 +28,7 @@ FEATURE_ENGINEER_REGISTRY: dict[ModelName, Type[BaseFeatureEngineer]] = {
     "FSRS-4.5": FSRSFeatureEngineer,
     "FSRS-5": FSRSFeatureEngineer,
     "FSRS-6": FSRSFeatureEngineer,
+    "FSRS-7": FSRSFeatureEngineer,
     "FSRS-rs": FSRSFeatureEngineer,
     "FSRS-6-one-step": FSRSOneStepFeatureEngineer,
     "RNN": FSRSFeatureEngineer,
--- a/models/__init__.py
+++ b/models/__init__.py
@@ -33,6 +33,7 @@ __all__ = [
     "FSRS4dot5",
     "FSRS5",
     "FSRS6",
+    "FSRS7",
     "FSRS_one_step",
     "FSRSRsBackend",
     "RNN",
--- a/models/__init__.py
+++ b/models/__init__.py
@@ -6,6 +6,7 @@ from .fsrs_v4 import FSRS4
 from .fsrs_v4dot5 import FSRS4dot5
 from .fsrs_v5 import FSRS5
 from .fsrs_v6 import FSRS6
+from .fsrs_v7 import FSRS7
 from .fsrs_v6_one_step import FSRS_one_step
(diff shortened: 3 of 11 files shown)


Expertium:
@L-M-Sherlock nope, I can't get it to work: the number of reviews is different and I don't know why.
When I run it using my (now outdated) benchmark code: "user": 1, "size": 16395
When I run it using the new benchmark code: "user": 1, "size": 11235

I use `python script.py --algo FSRS-7 --recency --secs --processes 20` for the new one and `python other.py --algo FSRS-7 --recency --secs --processes 20` for the old one.

Ah, nvm, I found a way to fix it: explicitly pass --short, like this: `python script.py --algo FSRS-7 --recency --short --secs --processes 20`. Now sizes do match.

Expertium:
Alright, so here's what's new in this version:
1) The total number of parameters is now 35.
2) The forgetting curve now uses 8 parameters.
3) `mean_reversion` now uses hard-coded values.
4) There are now 4 S-related formulas: stability_after_success_long_term, stability_after_success_short_term, stability_after_failure_long_term, stability_after_failure_short_term + a blending function for calculating an interval-length-dependent weighted average of short-term change in S and long-term change in S.
5) S can no longer be interpreted as "amount of time it takes for R to drop from 100% to 90%" (let's call it S90), so S90 will have to be calculated separately. Now S = a number used internally by FSRS, S90=the usual interpretation. Anki should display S90, which cannot be calculated analytically and has to be calculated numerically.
6) I changed the S0 inter/extrapolation method, the old one wasn't very good when S0 values could differ by orders of magnitude, like S0(Again) being 0.01 and S0(Easy) being 50.
7) Pretrain (or "parameter initialization") now supports initializing the forgetting curve parameters as well by iterating over several different candidates:
```
        initial_forgetting_curve_params = [
            self.init_w[-8:],
        ]
```
You can add more lists of 8 parameters. That way slightly better starting params can be found. Though I haven't found more starting forgetting curve parameters to use here that would improve the final log loss (after the entire optimization), but maybe you will find some.
8) Optimizer config:
```
    n_epoch: int = 8
    batch_size: int = 1024
    lr: float = 2e-2
    betas: tuple = (0.8, 0.85)  # this is for Adam, default is (0.9, 0.999)
```
This requires changing `get_optimizer` in a few places to accept custom betas. We can further reduce log loss by around 0.0010 by using NAdam instead of Adam and using 4 different values of lr for S0, params for S-related formulas, params for D-related formulas and params for the forgetting curve. But while that is straightforward to do in Pytorch, it would require a lot of extra effort to do so in Burn, so it's up to you whether you care enough to implement those optimizer changes.

I can also share the code that I used to obtain default parameters in DMs.

Expertium:
I'm currently doing more experiments to figure out good parameter ranges for the forgetting curve parameters. Take a look at this, I think it's an interesting approach, and once this PR is merged we can benchmark it among other things like default parameters only, S0, preset, etc.
[script_lstm_as_teacher.py](https://github.com/user-attachments/files/24845435/script_lstm_as_teacher.py)

Expertium:
Alright, so FSRS-7 is basically done. But I'm still experimenting with optimizing it, including the LSTM-as-teacher approach.

Expertium:
@L-M-Sherlock I'm still working on LSTM, but FSRS-7 itself is ready

Expertium:
@L-M-Sherlock should I open a new PR for GRU as a replacement of LSTM? IMO think you should merge this first

Expertium:
@L-M-Sherlock https://discord.com/channels/368267295601983490/1347982145418694747/1482103283290996857
https://discord.com/channels/368267295601983490/1347982145418694747/1482104928095834262
Seems like I still have to do more work. Though the difference in log loss is fairly small, so maybe it's fine to not care

PR #294, opened 2026-03-18
new result

asukaminato0721:


asukaminato0721:
note: someone add a new parameter, but it will need update all the result, too lazy to do it. so the table has problem now.

PR #328, opened 2026-05-18
Plot title now contains "With same-day reviews" or "No same-day reviews"

Expertium:


Expertium:
@L-M-Sherlock I don't think lint checks failing are related, this PR is ready to be merged

PR #335, opened 2026-05-26
Add FSRS-7 ordinal benchmark experiment

JSchoreels:
Adds an experimental FSRS-7-ordinal model and first 100-user benchmark results for comparison against plain FSRS-7.

@Expertium


JSchoreels:
## How to run and test

Run the new ordinal experiment for the first 100 users:

```bash
uv run script.py \
  --algo FSRS-7-ordinal \
  --short \
  --secs \
  --equalize_test_with_non_secs \
  --max-user-id 100 \
  --processes 8
```

Run the focused unit tests:

```bash
python3 -B -m unittest discover -s tests
```

Compare the ordinal result against plain FSRS-7 on the same users with `evaluate.py`:

```bash
python3 -B evaluate.py --fast \
  --models \
    FSRS-7-ordinal-short-secs-equalize_test_with_non_secs \
    FSRS-7-short-secs-equalize_test_with_non_secs \
  --max-user-id 100
```

The committed first-run result is:

```text
result/FSRS-7-ordinal-short-secs-equalize_test_with_non_secs.jsonl
```

Important evaluation note: the reported `.jsonl` `LogLoss` is still pure binary recall log loss from `R`. The auxiliary ordinal button likelihood is only used during training and is explicitly excluded from trainer eval/result metrics.

Expertium:
Thanks, I'll take a closer look tomorrow. Feel free to try different values of `ordinal_loss_weight`, including 0 just to ensure that it's equivalent to FSRS-7 before the PR

PR #337, opened 2026-06-06
perf: ~1.3-1.6x faster Trainer/data-pipeline path, bit-for-bit identical outputs

Expertium:
Written by Claude

## Summary
Speeds up the shared path used by all `Trainer`-class algorithms (FSRS-v*, DASH, HLR, …)
by ~1.3–1.6× while keeping outputs **bit-for-bit identical** (per-user `parameters` +
`LogLoss`, in practice all metrics).

Verified on the 20-user test sample (incl. the largest users) across 10 configs —
FSRS-7 default/recency/deck/preset/equalize, DASH, DASH `--short`, HLR, HLR `--short --secs`
all bit-identical; LogisticRegression within tolerance of its own run-to-run RNG /
fused-AdamW non-determinism (two unmodified runs differ by the same amount).

## Changes (each bit-exact)
- **LPT scheduling** (`script.py`): process the largest users first so a worker isn't stuck
  on a 1M-review user while others idle (shrinks the makespan tail). Order never affects
  per-user results (training shuffles via a private generator), and `sort_jsonl` re-sorts the
  output. Uses `user_order.py`, a precomputed user-id list in descending review-count order.
- **`rmse_matrix`** (`utils.py`): its three binning lambdas run over integer-valued columns,
  so memoize them over `Series.unique()` — identical results, transcendental math evaluated
  once per distinct value instead of once per row.
- **`create_features` dedup** (`features/base.py`): reuse the per-card history lists already
  computed in `_compute_histories` instead of recomputing the same `groupby/cum_concat` in
  `get_history_lists`.
- **`create_features` cum_concat** (`features/base.py`): replace 3× `groupby('card_id')
  .apply(cum_concat)` with a numpy-block iteration (numpy-scalar-preserving, so the history
  strings and tensors are identical), removing the pandas per-group `apply` overhead.
- **cards/decks read** (`data_loader.py`): read the direct hive-partition path
  (`cards/user_id=N`) instead of `read_parquet(..., filters=...)`, which re-discovers all
  ~3000 partitions on every call (~700 ms → ~3 ms); falls back to the filtered read for users
  with no partition dir.
- **FSRS tensor build** (`features/fsrs_engineer.py`): build one `[k, 2]` tensor per card and
  take prefix views instead of one `torch.tensor()` per review — same float32 values, and the
  intermediate column drops O(k²) → O(k) per card (memory win on huge users).
- **FSRS-7 constant hoists** (`models/fsrs_v7.py`): build the L2 `sigma`, `[7,16]` index, and
  zero-penalty tensors once in `__init__` instead of rebuilding them every
  `batch_process` / `stability_after_review`.

## Numbers
50-user size-stratified sample, `--default` (isolates the data pipeline): WALL **1.31×**,
makespan **1.35×**, total per-user work **~1.44×**; `create_features` alone **1.63×**. Trained
configs gain less per-user (training dominates) but still benefit from LPT's makespan packing.

The timing instrumentation in `script.py`/`utils.py` is gated by `SRSB_TIMING` and is
output-neutral (stderr only).


features/base.py
```diff
@@ -317,6 +350,12 @@ def get_history_lists(self, df: pd.DataFrame) -> Tuple[pd.Series, pd.Series]:
         Returns:
             Tuple of (time_history_list, rating_history_list)
         """
+        # Reuse the lists already computed in _compute_histories when available
+        # (set on this per-user engineer instance), avoiding a redundant recompute.
+        cached_t = getattr(self, "_cached_t_history_list", None)
```
asukaminato0721: https://docs.python.org/3/library/functools.html#functools.cached_property ？

features/base.py
```diff
@@ -138,7 +171,7 @@ def _compute_histories(self, df: pd.DataFrame) -> pd.DataFrame:
         return df
 
     def _compute_last_rating(
-        self, t_history_list: pd.Series, r_history_list: pd.Series
+        self, t_history_list: list, r_history_list: list
```
asukaminato0721: what is list[T] 's T

features/base.py
Expertium: Written by Claude

Thanks! I cleaned this up, but went with explicit cache attributes (initialized to `None` in `__init__`) rather than `functools.cached_property`. The history lists aren't lazily computed on first access — they're an eager byproduct of `_compute_histories` (which has to run anyway to set `last_rating` and the history strings) and they depend on the `df` argument, so a zero-arg `cached_property` on `self` doesn't model it cleanly. The explicit `self._cached_*` attrs make the "compute once in `_compute_histories`, reuse in `get_history_lists`" intent clear and drop the `getattr`. Happy to switch to `cached_property` if you'd prefer it.

features/base.py
Expertium: Written by Claude

Specified it — added a `HistoryLists = list[list[list[float]]]` alias and used it on `_cumulative_lists_by_card`, `_compute_last_rating`, and `_set_time_histories`. The nesting is: outer = cards, middle = reviews within a card, inner = the cumulative prefix of values up to (and including) that review. The innermost values are numpy scalars (`np.float64` for intervals, `np.int64` for ratings), so `float` is the closest single annotation.

PR #342, opened 2026-07-07
Add unit tests for rmse_matrix (RMSE(bins))

chrislongros:
Adds unit tests for `rmse_matrix`which currently has no test coverage.

Five cases, with hand-computed expected values:

- `test_calibrated`: a single bin with `mean(p) == mean(y)` returns 0
- `test_uniform_gap`: a constant per-bin gap propagates to the RMSE
- `test_unweighted_bins`: two bins combine as a root-mean-square
- `test_weighted_bins`: the `weights` column is summed per bin and used as `sample_weight`
- `test_lapse_zero_bin`: the `rmse_bins_lapse == 0` special case bins separately from the log-quantised values

CI currently runs only `ruff`, so these do not execute automatically. Happy to add a `pytest` job if that would be welcome.


asukaminato0721:
run `uv add pytest --dev` locally first.

then add a job :)

chrislongros:
@asukaminato0721 
Done. 
`uv sync --dev` pulls ~6.6 GB, mostly CUDA ...

asukaminato0721 (approved):

L-M-Sherlock (approved):

PR #348, opened 2026-08-29
FSRS-7 + GRU re-benchmark and other fixes

Expertium:
* FSRS-7: number of parameters changed to 34, better metrics
* new `--compile` flag, for speed. The committed FSRS-7 results were produced with `--compile`. Re-running in eager mode gives small differences. avg |Δ Log Loss| per user is ~7e-6 based on a sample of 50 users.
* models/fsrs_v6.py: forgetting curve moved to log space for `--compile` stability
* GRU: re-tuned hyperparameters from the Optuna search and re-benchmarked configs for both tables
* Removed stale GRU-P references
* Strict CLI parsing: raise an error on an unknown flag instead of silently skipping it
* `sort_jsonl()` made safe from crashes/interruptions corrupting the whole file
* Memory fix: script.py held on to the list of futures until the run finished, and a future never releases its result once computed, so every user's output stayed in RAM to the end (~3.6 GB on a `--raw` run). `as_completed()` already drops its own references, so dropping ours is enough. 14x lower peak memory, identical output
* Added per-user `time_ms` to result files. Machine-specific, only useful for *relative* comparisons (large vs small users, or FSRS-7 vs FSRS-6)
* evaluate.py now reproduces the README tables directly, with no manual editing (except for FSRS-7 deck rows). Rows sorted by Log Loss, bolding is applied automatically, MBE column commented out. Table header is now consistent (Algorithm, not Model)
* sped up BCa bootstrapping in evaluate.py by ~19x
* reptile_*.py moved into a reptile/ folder
* RWKV naming: RWKV-P → RWKV-Instant, RWKV → RWKV-Curve
* Significance/superiority plots regenerated

Not included:

* `FSRS-7-*-deck.jsonl` — too large for GitHub


Expertium:
@asukaminato0721 I also suggest merging https://github.com/open-spaced-repetition/fsrs-rs/pull/426

Checks are failing because of return types and pyrefly

Issue #354, opened 2026-09-12
New model submission


andersschill:
A 7-parameter model that is only marginally behind FSRS-7 in terms of accuracy.

Paper and code available at:
https://github.com/andersschill/memory-model-benchmark-for-spaced-repetition

Expertium:
@L-M-Sherlock wanna add it to the benchmark? I have no objections.

Expertium:
Btw, @andersschill I would appreciate a .pdf of your paper 🙏

andersschill:
Thanks. You should be able to get a pdf from here:
https://doi.org/10.5281/zenodo.22727105

L-M-Sherlock:
PR is welcome.

PR #355, opened 2026-09-13
Add SBD, a 7-parameter memory model

Expertium:
Model submission from #354: SBD by @andersschill ([paper](https://doi.org/10.5281/zenodo.22727105), [original code](https://github.com/andersschill/memory-model-benchmark-for-spaced-repetition/blob/main/models/sbd.py)). This PR adds the model only. @L-M-Sherlock @asukaminato0721, could one of you benchmark it on the 10k dataset? It is CPU-only; run it with `--recency`, with and without `--short --secs`.

**What's in it**
- `models/sbd.py`: the 7-parameter recurrence, copied from the original line for line (same safe log/exp, same clamps, same order). Only the tensor layout is adapted to srs-benchmark's `[seq_len, batch, 2]` sequences. The module docstring has the closed-form formulas.
- Training follows the original, not the Adam `Trainer`: full-batch L-BFGS (strong Wolfe, ≤200 iterations) from the pretrained defaults, with an L2 pull toward them (`lam = 1e-3 · 2000 / n_targets`, per-parameter sigmas from the original), reverting to the defaults if the fit ends worse. `_fit_trainable_weights` dispatches SBD to `model.fit()`.
- `--recency` reuses the FSRS-7 branch, because SBD's recency formula is the same one (`0.0667 + 0.9333·x^11.25`). The original always trains with recency weighting, so SBD is benchmarked with `--recency`.
- Registration in `config.py`, `models/__init__.py`, `models/model_factory.py`, `features/factory.py` (FSRS feature engineer: elapsed days and rating).

**Verification against the original code**
- Predictions agree to max |diff| 6e-8 (3.6e-7 on padded batches, float32).
- The closed-form formulas in the docstring agree with the original to 1e-6 over 300 random review sequences.
- End-to-end smoke test on 3 users (fit + 5-fold evaluation).

**Known differences from the original setup**
- `FIT_BATCH_SIZE = 8192` instead of 1024. The objective is a sum over all examples, so the batch size only changes float summation order. On a 65.8k-row fold the parameters agreed within 6e-4 and the fit was 4.4× faster.
- Examples with a history longer than 64 reviews are dropped by srs-benchmark's `BatchDataset`, as for the FSRS models here.

**Early numbers** from a partial local run (the 187 largest users, `--short --secs --recency`, size-weighted): SBD LogLoss 0.3287 vs FSRS-7 recency 0.3236, RMSE(bins) 0.0405 vs 0.0357, AUC 0.734 vs 0.740; SBD is behind on 167/187 users. The gap is in line with the original repo's table (0.3279 vs 0.3237).

@andersschill could you check that the port matches your implementation, in particular the L-BFGS fit and the L2 anchors? If anything is off, tell me and I'll fix it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


L-M-Sherlock:
The 10k-user benchmark is underway. The `--recency` run is in progress, with `--short --secs --recency` planned next. SBD's full-batch L-BFGS fitting is memory-intensive, so I'm using 4 CPU workers to keep memory usage manageable. It will take some time to complete both runs.

L-M-Sherlock:
## Benchmark progress update (2026-09-15)

The **day-level `--recency` run is complete and validated**:

- 9,999/9,999 eligible users (user 4371 has no eligible day-level data, matching the existing benchmark)
- 349,923,850 evaluation reviews
- SBD: **Log Loss 0.3484 ± 0.0043**, **RMSE(bins) 0.06215 ± 0.00098**, **AUC 0.7010 ± 0.0020** (user-weighted mean ± 99% BCa CI)
- For reference, FSRS-7 recency is 0.3370 / 0.0593 / 0.7220. SBD has lower per-user Log Loss for 1,364/9,999 users.
- Against the more input-comparable FSRS-6 recency run without same-day history, SBD is slightly better on all three means: 0.34835 vs 0.34984 Log Loss, 0.06215 vs 0.06410 RMSE(bins), and 0.70097 vs 0.69571 AUC. SBD wins per-user Log Loss on 5,489/9,999 users (54.90%).

The **`--short --secs --recency` run is in progress**. A frozen interim snapshot contains 103 completed users and 59,904,465 evaluation reviews (11.54% of all same-day evaluation rows). Users are processed largest-first, so this is a deliberately biased large-user subset and is **not** a final estimate.

| Model | Log Loss ↓ | RMSE(bins) ↓ | AUC ↑ |
| --- | ---: | ---: | ---: |
| RWKV-Instant | 0.27383 | 0.01315 | 0.83772 |
| RWKV-Curve | 0.30113 | 0.02864 | 0.78927 |
| LSTM | 0.31437 | 0.03057 | 0.75921 |
| GRU | 0.31503 | 0.03063 | 0.75606 |
| FSRS-7 recency | 0.31946 | 0.03660 | 0.74453 |
| FSRS-7 | 0.32225 | 0.04209 | 0.73710 |
| Logistic Regression | 0.32347 | 0.04109 | 0.73230 |
| **SBD** | **0.32504** | **0.04164** | **0.73802** |
| MOVING-AVG | 0.33538 | 0.06866 | 0.70649 |
| FSRS-6 recency | 0.36274 | 0.07649 | 0.68175 |
| FSRS-6 | 0.36559 | 0.07854 | 0.67846 |
| FSRS-5 | 0.42987 | 0.08976 | 0.66237 |

All interim comparisons use exactly the same 103 users and matching per-user evaluation sizes. On this subset, SBD beats FSRS-7 recency for 14/103 users on Log Loss, 28/103 on RMSE(bins), and 15/103 on AUC. It is close to plain FSRS-7/Logistic Regression, but remains behind FSRS-7 recency and the neural models.

The run is active under a persistent user service. Full-batch L-BFGS is memory-bound: current concurrency is about four large users, service memory is typically ~15–21 GiB with a measured peak of 25.2 GiB, swap usage is zero, and there have been no OOM events. Concurrency will increase automatically as user sizes decrease. I will post the final same-day result after all 10,000 users finish and the coverage/row-count validation passes.

andersschill:
Correct me if I'm wrong. I'm not very familiar with your codebase. But it seems to me like results for the "without same-day reviews" table should be produced with:
`--short --secs --recency --equalize_test_with_non_secs`

And "with same-day reviews":
`--short --secs --recency`

It sounds like the completed run just had `--recency`, which would mean SBD got integer intervals and that same-day reviews were removed, both as history and as training targets. Its default weights were fit on fractional intervals with same-day reviews included so it matters a lot.

Expertium:
For FSRS-7:
With same-day reviews `--short --secs --recency`.
Without same-day reviews `--short --secs --recency --equalize_test_with_non_secs`.
FSRS-7 is intended to **always** be used with fractional interval lengths, **never** with integer interval lengths. If the same is true for SBD, then the same flags should be used for SBD, yes.
@L-M-Sherlock

andersschill:
Perfect. Please do the same for SBD.

L-M-Sherlock:
## Final SBD benchmark results

The benchmark has been completed and validated on PR commit `6d50fdd778c4`.

Both runs use the SBD implementation from this PR: seven trainable parameters, the original full-batch L-BFGS fit (up to 200 strong-Wolfe iterations), recency weighting, five chronological folds, and a 64-review history cap.

### Without same-day reviews

Command:

```bash
uv run python script.py --algo SBD --short --secs --recency --equalize_test_with_non_secs
```

This keeps fractional intervals and same-day history during training while equalizing the evaluation rows with the established non-seconds table.

- 9,999 eligible users (user 4371 has no eligible day-level evaluation data)
- 349,923,850 evaluation reviews
- Log Loss: **0.34096 ± 0.0042**
- RMSE(bins): **0.06018 ± 0.00097**
- AUC: **0.71524 ± 0.0021**

Against FSRS-7 recency on the same users and evaluation sizes, SBD has lower Log Loss for 2,703/9,999 users (27.03%). The mean paired Log Loss difference is +0.00400 (SBD − FSRS-7 recency).

### With same-day reviews

Command:

```bash
uv run python script.py --algo SBD --short --secs --recency
```

- 10,000 users
- 519,296,315 evaluation reviews
- Log Loss: **0.32179 ± 0.0040**
- RMSE(bins): **0.05862 ± 0.00083**
- AUC: **0.74454 ± 0.0018**

Against FSRS-7 recency on the same users and evaluation sizes, SBD has lower Log Loss for 2,308/10,000 users (23.08%). The mean paired Log Loss difference is +0.00397 (SBD − FSRS-7 recency).

For context, with same-day reviews the README table reports FSRS-7 recency at 0.3178 / 0.05715 / 0.7522, while SBD is 0.3218 / 0.05862 / 0.7445. SBD is close to plain FSRS-7 and Logistic Regression, and uses only 7 parameters.

The per-user JSONL results, generated plots, validation report, and benchmark runner are included in this commit. Verification passed with `5 passed` and Ruff checks passed.

Expertium:
Nice. If @andersschill doesn't see any problems, feel free to merge this PR

andersschill:
Looks good. Thank you.

